PR #3, opened 2016-02-18
wrote Dockerfile that actually works

KekSfabrik:
Your Dockerfile didn't really do anything so I wrote one that works


wrouesnel:
Sorry to have let this languish. Thanks for the pull request but it's obsoleted by 054d825 which builds a compact dockerfile by chaining the build processes together (and also brings in vendoring of the dependencies). This should allow anyone to build the docker image locally, which I've now uploaded to the docker hub.

KekSfabrik:
Even better - thanks :100:

PR #5, opened 2016-05-06
Add Dockerfile

marcusmartins:
Bypass the custom build script and use a Dockerfile to build the service. 

The docker file is based on Alpine and it's much smaller image than golang:1.6-wheezy

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,13 @@
+FROM golang:1.6-alpine
+
+RUN apk --update add make
+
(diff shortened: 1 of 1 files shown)


wrouesnel:
The main point of the build script is that the binary only is what ends up in the final docker-image. The problem with the alpine and other containers is you end up with a multilayered image which contains the entire golang build infrastructure as well.

wrouesnel:
The new docker build changes + CI of the image with travis take care of this.

PR #6, opened 2016-05-06
Handle `pg_runtime_variable` return values

marcusmartins:
When querying postgres `pg_runtime_variable`, postgres might return -1
for `max_standby_archive_delay` and `max_standby_streaming_delay`

### DIFF
--- a/postgres_exporter.go
+++ b/postgres_exporter.go
@@ -286,6 +286,10 @@ func makeDescMap(metricMaps map[string]map[string]ColumnMapping) map[string]Metr
 							return math.NaN(), false
 						}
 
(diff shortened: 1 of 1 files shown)


wrouesnel:
Should this perhaps be a `math.Inf(1)` for these values since they both technically imply infinite wait times, and Prometheus supports that representation?

wrouesnel:
So I let this languish far too long, but the answer is kind of obvious based on my prior experience - `math.Inf(1)` in Prometheus tends to break graphs and in this particular case the lead up to  a -1 is not steadily increasing wait times AFAIK. Merging.

PR #9, opened 2016-05-27
Added application_name label to pg_xlog_location_diff gauge

kkoppel:
I have a PostgreSQL master with multiple replication slaves, which all come from the same source IP and have the same state usually.
Currently this results in the following output from postgres_exporter:
pg_stat_replication_pg_xlog_location_diff{client_addr="192.168.2.4",state="streaming"} NaN
pg_stat_replication_pg_xlog_location_diff{client_addr="192.168.2.4",state="streaming"} 48

and Prometheus only stores the first ("NaN") value. I added the application_name label to get Prometheus  to store both results.


wrouesnel:
Looks good and solves the problem. Merged. CI should automatically upload it.

kkoppel:
Looks like this change actually never made it to the postgres_exporter repository. Could you check what happened to it?

PR #14, opened 2016-07-20
Use tini as a minimal init for the docker image

### LINKED ISSUE #12: Huge number of open file descriptors after days of running
Hi Will,

We've been running pg exporters for the last 2 weeks, and we noticed that they leaked file descriptors.

Here's a list of some pg exporters running on our site, the 2nd column being the open fds on the docker daemon and names have been masked:
```
prometheus-postgres-xxxxxxxxx-replica  18959
prometheus-postgres-xxxxxxxxx-master   11800
prometheus-postgres-xxxxxxxxx-master   11780
prometheus-postgres-xxxxxxxxx-master   11796
prometheus-postgres-xxxxxxxxx-master   11790
prometheus-postgres-xxxxxxxxx-master   11782
prometheus-postgres-xxxxxxxxx-master   11810
prometheus-postgres-xxxxxxxxx-replica  2868
prometheus-postgres-xxxxxxxxx-replica  2870
prometheus-postgres-xxxxxxxxx-replica  2875
prometheus-postgres-xxxxxxxxx-replica  2872
prometheus-postgres-xxxxxxxxx-replica  2862
prometheus-postgres-xxxxxxxxx-replica  2875
```

As a result, this exposes a docker bug, where even destroying those containers, leaves fds deleted but still open on the docker daemon.

Here's a list of open fd count on our DB servers:
```
docker_fd_count                      125996
docker_fd_count                                373688
docker_fd_count                           484407
docker_fd_count                            520493
```

The last one is at half a million open fds, and it runs our masters.

`docker ps` just hangs indefinitely.

To reproduce this bug, just run an exporter and wait, open fds on docker daemon will keep growing.

Thank you

noushi:
This PR fixes #12 (so far).

It's based on alpine:3.4 to minimize image size.

Please notice that I only modified the docker make target, docker-build will need to be modified as well.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM alpine:3.4
+
+RUN apk add --update tini
+
+ADD postgres_exporter /
+
+EXPOSE 9113
+
+ENTRYPOINT ["/sbin/tini", "--", "/postgres_exporter"]
--- a/Makefile
+++ b/Makefile
@@ -11,9 +11,7 @@ postgres_exporter: $(GO_SRC)
 
 # Take a go build and turn it into a minimal container
 docker: postgres_exporter
-	tar -cf - postgres_exporter | docker import --change "EXPOSE 9113" \
-			--change 'ENTRYPOINT [ "/postgres_exporter" ]' \
-			- $(CONTAINER_NAME)
+	docker build -t $(CONTAINER_NAME) .
 
 vet:
 	go vet .


wrouesnel:
So I don't know what to do about this one because it's an issue I can't replicate anywhere else.

noushi:
Aside from the bug we experienced and solved using this PR, it's better to have both:
- a Linux environment one can use to enter a running container
- signal handling using an init-compatible entrypoint

Basically you have two ways of managing signals:
- you use an init-compatible entrypoint
- you handle the signals yourself in your entrypoint (postgres_exporter in our case)

That said, although Prometheus (server) handles signals itself (as shown [here](https://github.com/prometheus/prometheus/blob/0938661db93af4b8367a9ef331ad8f3e27da80fd/cmd/prometheus/main.go#L124-L137) and [here](https://github.com/prometheus/prometheus/blob/0938661db93af4b8367a9ef331ad8f3e27da80fd/cmd/prometheus/main.go#L190-L197)), it still uses a minimal base image, with regular Linux tools, like sh, ps, cat... 

If not for the bug fix, this PR serves as two feature enhancements:
- signal handling
- OS environment in image

wrouesnel:
So I'm going to close this for now, as the underlying issue I haven't been able to replicate. Closes #12 as well.

My preferred fix for this is to handle the correct signals in the `postgres_exporter` binary, which as far as I know does happen at the moment (in the sense that I have not been able to replicate the issue locally).

noushi:
Sure, thank you @wrouesnel .

PR #17, opened 2016-07-28
Parse strings to floats

jinty:
it appears that the SHOW queries were returning the "string" type

First go code I ever wrote...
fixes https://github.com/wrouesnel/postgres_exporter/issues/15


wrouesnel:
Looks good - merged and thanks!

PR #18, opened 2016-08-06
update to use common/log

tmc:


wrouesnel:
Thank you for the PR! Definitely something we want to do. Merged.

postgres_exporter.go
```diff
@@ -330,7 +327,7 @@ func dbToFloat64(t interface{}) (float64, bool) {
 	case string:
 		result, err := strconv.ParseFloat(v, 64)
 		if err != nil {
-			log.Println("Could not parse string:", err)
+			log.Infoln("Could not parse string:", err)
```
grobie: All of these should be `Errorln` (or maybe `Warnln`).

PR #20, opened 2016-09-06
Enable crossbuilds with Promu

deathowl:


wrouesnel:
What's this PR actually attempting to do? 

Also I've just committed a different upstream fix for the Makefile - thanks Docker for breaking that :)

deathowl:
it enables the application to build for different platforms like linux-amd64, linux-i386, windows via the prometheus builder utility called promu.

wrouesnel:
Closing this in favor of the final fate of #79 (which will probably be broken up into a few other changes).

PR #11, opened 2016-06-29
Feature queries from file

noushi:
Hi @wrouesnel ,

This PR adds support for custom queries in an external YAML file.

SHOW variables are not supported yet.

It should help close #1 .

I've deployed it yesterday and we're successfully getting metrics from our 54 Postgres replicas.

PR #21, opened 2016-09-07
Feature queries from file

ncabatoff:
This is basically the same as PR #11 from @noushi, plus an important bugfix.  I already filed a PR with that bugfix against noushi/postgres_exporter, but I decided to file this new PR directly to make it easier to incorporate (and so I wouldn't have to rely on two people accepting PRs.)


wrouesnel:
I've dilly-dallied long enough on this - it's a good concept that's clearly useful and builds well, so let's get it merged!

ncabatoff:
Excellent, thanks to you and @noushi.

PR #23, opened 2016-09-08
Add slot_name to pg_stat_replication

mcanevet:


wrouesnel:
This looks good (at the very least doesn't egreriously break the smoketests) so we'll merge it.

Reminds me that if anyone can figure out a way to setup a replicated postgres in travis for build time tests of the WAL replication, that'd be awesome.

PR #31, opened 2016-10-14
ADD more flag for don't use the environment variable

mperochon:
… a file for the postgres password


wrouesnel:
I don't think I want to merge this type of functionality in, because it can encourage bad practices (unencrypted files on disk). The environment variable usage works with most types of launch systems, but better can work with more secure storage mechanisms (i.e. launch vault, get env vars, etc.).

The same effect can be achieved by launching with docker using an environment file - `docker run --env-file=/path/to/my/file wrouesnel/postgres_exporter` or by using systemd and the `EnvironmentFile=` directive.

PR #32, opened 2016-10-26
add a VERSION file, similar to other prometheus exporters

erkki:
MLove Version to a separate file, similar to how other prometheus exporters do it (eg the node exporter).  This simplifies/unifies configuration management over installing different prometheus exporters.


wrouesnel:
The build system is currently based on git tags. I'm not sure what a VERSION file is actually bringing to the table?

erkki:
Uniformity with some of the upstream maintained exporters.

wrouesnel:
Closing this. For the exporter we use git tags.

PR #35, opened 2016-11-19
add pg_stat_statements metrics

tmc:
This incorporates pg_stat_statements in the example additional queries.


wrouesnel:
Could you rebase this? Also, do you think this would be a good candidate for being made an internal metric (how expensive is this query, basically)?

homelessnessbo:
Pay attention to this PR please, we also need this functionality in our monitoring system.

queries.yaml
```diff
@@ -98,3 +98,17 @@ pg_stat_user_tables:
...
+  metrics:
+    - name:
+        usage: "LABEL"
+        description: "Name of the schema that this table is in"
+    - owner:
+        usage: "LABEL"
+        description: "Owner of this database"
+    - size:
+        usage: "GAUGE"
+        description: "Owner of this database"
```
databus23: I think this description is a copy & paste typo

queries.yaml
```diff
@@ -98,3 +98,17 @@ pg_stat_user_tables:
     - autoanalyze_count:
         usage: "COUNTER"
         description: "Number of times this table has been analyzed by the autovacuum daemon"
+
+pg_database:
+  query: "SELECT d.datname AS name, pg_catalog.pg_get_userbyid(d.datdba) AS owner, CASE WHEN pg_catalog.has_database_privilege(d.datname, 'CONNECT') THEN pg_catalog.pg_database_size(d.datname) ELSE 0 END AS size FROM pg_catalog.pg_database d ORDER BY CASE WHEN pg_catalog.has_database_privilege(d.datname, 'CONNECT') THEN pg_catalog.pg_database_size(d.datname) ELSE NULL END DESC"
+  metrics:
+    - name:
+        usage: "LABEL"
+        description: "Name of the schema that this table is in"
```
databus23: And also doesn't really sound right. Should be "Name of the database"

queries.yaml
```diff
@@ -98,3 +98,17 @@ pg_stat_user_tables:
...
+    - owner:
+        usage: "LABEL"
+        description: "Owner of this database"
+    - size:
```
mattbostock: This should include the units, e.g. `size_bytes`.

sharmay:
FYI, requires  pg_stat_statements so to be loaded via shared_preload_library statement.

uspen:
May be so:

pg_stat_statements_all:
  query: "SELECT coalesce(d.datname, '~unknown') AS database, coalesce(r.rolname, '~unknown') AS user, query, calls, total_time, min_time, max_time, mean_time, stddev_time, rows, shared_blks_hit, shared_blks_read, shared_blks_dirtied, shared_blks_written, local_blks_hit, local_blks_read,   local_blks_dirtied, local_blks_written, temp_blks_read, temp_blks_written, blk_read_time, blk_write_time FROM pg_stat_statements s JOIN pg_roles r ON r.oid=s.userid JOIN pg_database d ON d.oid=s.dbid WHERE d.datname NOT IN ('postgres', 'template0', 'template1')"
  metrics:
    - database:
        usage: "LABEL"
        description: "Datname"
    - user:
        usage: "LABEL"
        description: "Role"
<SKIP>

?

gregscds:
This pull request says "add pg_database_size metric", but what it actually does is use a view from the pg_stat_statements extension.  I can't tell whether the votes for this pull were based on its description (database size) or its actual function (statement statistics)
.  
Supporting statements statistics correctly is a lot more work than just adding this query.   pg_stats_statements may not be loaded.  Also, that extension been actively in development where the fields available depends heavily on which version of PostgreSQL you're running.

I would suggest closing this pull because it can't just be coded and merged this easily.  Then a proper issue for pg_stats_statements should be opened to work out how to handle the tricky bit of version specific work.

wrouesnel:
@gregscds Thanks for the analysis. It does raise the bigger issue of how that `queries.yaml` file included in the repo should be supported. I suspect it needs to evolve towards being a bundle of examples that can be dropped in by users - since its perfect for site-specific extra metrics.

strangeman:
Got `fatal error: runtime: out of memory` with `queries.yaml` from that PR: https://pastebin.com/qVmrwYUj
BTW, with `queries.yaml` from the current master all working correctly.

tmc:
I've looped around at another job and now I want this again.

binakot:
What news with this request? 🐘

queries.yaml
```diff
@@ -122,3 +122,67 @@ pg_database:
...
+pg_stat_statements:
+  query: "SELECT query, calls, total_time, min_time, max_time, mean_time, stddev_time, rows, shared_blks_hit, shared_blks_read, shared_blks_dirtied, shared_blks_written, local_blks_hit, local_blks_read, local_blks_dirtied, local_blks_written, temp_blks_read, temp_blks_written, blk_read_time, blk_write_time FROM pg_stat_statements"
+  metrics:
+    - query:
+        usage: "LABEL"
+        description: "Query class"
+    - calls:
+        usage: "COUNTER"
+        description: "Number of times executed"
+    - total_time:
```
baryluk: This, and other `..._time` stuff. Shouldn't they all have units in the suffix of the metric? I.e. `pg_stat_statements_total_time_seconds` , so `SELECT ... , total_time / 1000 as total_time_seconds, ...`

See https://prometheus.io/docs/practices/naming/#metric-names and https://github.com/wrouesnel/postgres_exporter/issues/261

mbanck:
The `queryid` column is present in all supported versions of Postgres (i.e. in 9.4+), so maybe nowadays it would make sense to add it as `LABEL` in order to avoid issues like in https://github.com/wrouesnel/postgres_exporter/issues/219#issuecomment-426312030

wrouesnel:
Than you all for comments on this, this has lingered long enough. I've fixed the original issues and re-issued it as a new PR.

mbanck:
I still think exposing the `queryid` column is  a good idea, even more so now that 0.6.0+ requires Postgres 9.4+ anyway and it will always be present.

Otherwise, people risk getting "collected metric ... was collected before with the same name and label values" errors from Prometheus for the 'insufficient privileges' case as `query` content.

PR #42, opened 2016-11-29
Properly print the SQL server error string in the log message.

EdSchouten:
On a Debian + PostgreSQL 9.4 system I had a hard time getting the
exporter to work. It looks like I was running into some authentication
issues that weren't logged by sql.Open(), but by sql.DB.QueryRow().

This change extends the warning generated by the call to
checkMapVersions() to also print the error message generated by
sql.DB.QueryRow().


wrouesnel:
Thanks!

PR #48, opened 2017-02-07
Using a connection pool of size 1 instead of reestablishing connection all the time

matevarga:
https://github.com/wrouesnel/postgres_exporter/issues/47


wrouesnel:
Thanks! Approved!

wrouesnel:
I've also cut a `v0.1.2` off this release so it should be available for use with docker/binaries right away.

PR #51, opened 2017-02-15
Improved error message

jtilander:
Made some changes so that the log messages are clearer when one have messed up the connection string. 

Fixed building of the container on a non linux native machine (e.g. Mac)


wrouesnel:
Looks good!

PR #63, opened 2017-04-10
Use latest minor version of Go 1.7

cmur2:
I am interested in the postgres_exporter releases built with any Go > 1.7.0 since that should fix the issue https://github.com/golang/go/issues/16739 in Go.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ services:
 - docker
 language: go
 go:
-- '1.7'
+- '1.7.5'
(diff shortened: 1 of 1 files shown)


cmur2:
Thanks! @wrouesnel do you plan to do a new release in the nearer future?

wrouesnel:
I'll point release this later tonight.

wrouesnel:
`v0.1.3` has been pushed and should be available soon.

PR #65, opened 2017-04-17
Handle pg_settings names with '.'

### LINKED ISSUE #64: Amazon RDS pg_settings
On Amazon's RDS there are additional settings that are getting picked up via pg_settings which are eventually causing an error due to the name of the setting:
The problem settings are:
```
rds.rds_superuser_reserved_connections | 2         |          | Sets the number of connection slots reserved for rds_superusers.    | integer
rds.restrict_logical_slot_creation     | off       |          | Restrict logical slot creation over streaming replication protocol. | bool
```

These two rows ultimately end up causing a panic:
```
panic: descriptor Desc{fqName: "pg_settings_rds.rds_superuser_reserved_connections", help: "Sets the number of connection slots reserved for rds_superusers.", constLabels: {}, variableLabels: []} is invalid: "pg_settings_rds.rds_superuser_reserved_connections" is not a valid metric name
```

I'm happy to make a patch for this, but I'm not sure what the preferred approach would be. A simple fix would be to substitute the `.` with an `_` in the `pg_settings.name` field. We could also just ignore any rows starting with `rds.`.

iterion:
Fixes #64

### DIFF
--- a/pg_setting.go
+++ b/pg_setting.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"strconv"
+	"strings"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/log"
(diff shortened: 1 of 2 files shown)


wrouesnel:
Superb - merged. I think with that I'll do another release.

PR #66, opened 2017-04-18
Remove pg_settings query examples

iterion:
These fields are now automatically scraped so they will conflict if they are accidentally specified.

Some additional cleanup for the v0.2.0 release.


mattbostock:
👍 LGTM.

PR #82, opened 2017-06-10
bash style fixes

tamsky:


wrouesnel:
I'm all for this PR, but I'd be happier if there was an associated linter or tool references which is making/can enforce these recommendations.

tamsky:
I was doing it by hand.

Since you mentioned it, and I agreed it's a good idea.
I fired up https://github.com/koalaman/shellcheck and have started following its suggestions.

I'll report back shortly with a green build and a make target.

tamsky:
@wrouesnel this is ready for review.

I've added the shellcheck linter to the Makefile.
I've updated shell scripts to address findings and harmonize styles across scripts.

tamsky:
> Could you rebase and squash this down to a few logical commits?

Rebased.
Also, "Allow edits from maintainers." is enabled for this PR.

wrouesnel:
Tried merging this but the scripts have deviated a lot. The improvements will be incorporated as a fresh set of changes.

PR #83, opened 2017-06-11
ignore this while I work out the failing travis build.

tamsky:
start with smaller changes


tamsky:
I had inadvertently inverted the logic on two similar expressions during my refactor.

PR #103, opened 2017-09-12
Removed unnecessary complexity in readme.md

einhverfr:
In PostgreSQL, views run as the defining user, meaning you don't 
need a view wrapping a function (you can just have a view).  This 
improves readability, maintainability, performance, and clarity.


cbandy:
I've seen the `security definer`-function approach to monitoring used by many projects.&sup1; Does directly using a view have any other side-effects, like maybe allowing writes?

&sup1; https://github.com/pganalyze/collector#setting-up-a-restricted-monitoring-user

einhverfr:
These views are not updatable.  It is true that depending on default permissions views wrapped in ways like this may or may not be updatable but in my experience with PostgreSQL since 6.5 I would note that by default people don't get write permissions to tables or views.

Additionally these are built on system views which are not updatable.  So in these cases, write permission is not at all an issue.   If the underlying relation was updatable then you could grant access to other users to write.

wrouesnel:
I think the main thing I need to do is add some integration testing around non-superuser monitoring. That way we can have some confidence worth changes like this.

wrouesnel:
I can't see a reason not to merge this, so let's merge it!

PR #106, opened 2017-09-19
Add ability to get DSN from multiple Cloud Foundry service bindings

ihrigb:
This enables the service (if `DATA_SOURCE_NAME` is not specified) to try to parse `VCAP_SERVICES` in order to get all bound postgres services.
If it finds some, it will retrieve the URIs and enable multiple exporters. This is while `userQueriesPath`-file will be applied to each of them and it will search for a file with name `<cf-service-instance-name>.yml`, which will be individually added.


ihrigb:
Need to do some rework on this.

PR #107, opened 2017-09-26
Add the ability to connect to multiple Cloud Foundry datasource bindi…

ihrigb:
…ngs.

This needs to be enabled via --use-cf-bindings.


wrouesnel:
This is a fairly significant change which I need to figure out the ramifications of. It's flirting pretty closely to adding blackbox_exporter style support, where Prometheus would instead hit a `/probe` endpoint and send the relevant polling details.

PR #112, opened 2017-10-08
be tolerant when parsing the postgres version

rgl:


wrouesnel:
We don't particularly want to be more tolerant about version mismatches, since it can lead to subtly incorrect metrics (not a situation you want to be in).

The correct fix here is to handle the 10.x version transition with the semver library properly, which will probably involve just appending a `.0` so it looks right. I'll upload a patch soon.

rgl:
I'm not seeing why ParseTolerant is not a good fit, since it pretty much does that.

wrouesnel:
You're right it is! I have an alternate fix sitting on another PR (adds a better regex check) but the integration tests are not reporting full metrics for postgres 10 yet so more work to do.

PR #137, opened 2017-11-27
fix invisibility of "/metrics"

nkonev:
see https://github.com/wrouesnel/postgres_exporter/issues/136


nkonev:
@wrouesnel please merge it ;)
I just put html template code after kingpin parsing

nkonev:
Thank you

PR #141, opened 2017-12-08
Add the way to prefix metrics' names using flag "--extend.metric-prefix"

dangusev:
Hi! I added some simple functionality to prefix each metric's name. It should prevent the clashing of metrics from several sources in one Prometheus storage.


dangusev:
Actually, I found another way to solve my problem with clashing names via exporter settings. That makes this PR useless, so I close it).

PR #149, opened 2018-01-22
In docker example both containers should run on the host net

schnipseljagd:
IMHO both container have to run in the host net.
Or did I get something wrong here?


wrouesnel:
No that's a good catch, but we don't need `-p` then either.

schnipseljagd:
Right, didn't see that.

PR #158, opened 2018-02-28
Allow default metrics to be disabled in order to support wider range of postgres versions

hsun-cnnxty:
I would like to use this exporter to collect metrics from older versions of Postgres and Greenplum (a product derived from Postgres).  To do that,  the default metrics and query overrides need to be disabled so that they don't cause incompatible issues.  After that, a set of version specific metrics can be defined and imported using existing mechanism.  This PR implements a command-line flag to disable the default metrics. It has been tested to export custom metrics from Greenplum.

-thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ postgres_exporter_integration_test
 *-stamp
 .idea
 *.iml
+.vscode
 cover.out
 cover.*.out
 .coverage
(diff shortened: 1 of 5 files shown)


srini3463:
@hsun-cnnxty  Please let me know if it supports to get the metrics from Greenplum server.

Thanks.

srini3463:
@sunh11373  Thank you so much for your response. It would be great if you could provide some sample queries to get basic metrics. I am new to Greenplum and do not have any idea about it.

Appreciate your help.

Thanks.

PR #179, opened 2018-04-13
Seiferbb issue 173

Seiferbb:
I found the unit of postgres_exporter/pg_setting.go should be fix too


wrouesnel:
Superseded by #177

PR #189, opened 2018-05-22
FreeBSD compilation target

thomersch:
Added FreeBSD as a building target. Tested cross-compiling on local machine and moved binary to FreeBSD target, which ran it without problems.


wrouesnel:
Fixed the tests so lets see if it rebuilds now....

thomersch:
Tests have passed :) Can we get this merged?

PR #191, opened 2018-05-22
URL encode username/password fields

### LINKED ISSUE #188: Usernames/passwords supplied via DATA_SOURCE_PASS_[FILE] are not URL encoded
Some of our passwords contain characters that are not valid in URLs; they need to be escaped prior to formatting: https://github.com/wrouesnel/postgres_exporter/blob/2db74865b1ec8758cc798835cc109ac0f5c98dfc/cmd/postgres_exporter/postgres_exporter.go#L1090

NB it is not possible to pre-encode the password as a workaround, as it is sourced from a kubernetes secret that is observed by other components which expect it to be in its literal form.

awh:
Fixes #188.

It occurs to me that you might have a concern about this changing behaviour in the case where existing users have worked around the problem by URL encoding the password themselves before supplying it (it will double encode resulting in authentication failure). If that's the case, I can change the PR so that it only applies the encoding if it detects that it is necessary; this should leave existing unaffected users unaffected, but also fix the problem for those that are.

### DIFF
--- a/cmd/postgres_exporter/postgres_exporter.go
+++ b/cmd/postgres_exporter/postgres_exporter.go
@@ -1086,8 +1086,9 @@ func getDataSource() string {
 			pass = os.Getenv("DATA_SOURCE_PASS")
 		}
 
+		ui := url.UserPassword(user, pass).String()
 		uri := os.Getenv("DATA_SOURCE_URI")
-		dsn = "postgresql://" + user + ":" + pass + "@" + uri
+		dsn = "postgresql://" + ui + "@" + uri
 	}
 
 	return dsn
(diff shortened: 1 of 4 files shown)


awh:
It doesn't look to me like the CI error was introduced by my change:

```
The command "go run mage.go -v all" exited with 1.
0.01s$ $HOME/gopath/bin/goveralls -coverprofile=cover.out -service=travis-ci
Error parsing coverage: open cover.out: no such file or directory
```

wrouesnel:
`ping not found` is certainly a new one to me from that docker build. I guess some package deps changed upstream in ubuntu.

awh:
Ah thanks!

PR #198, opened 2018-06-19
Fix comments about DATA_SOURCE_* environment variable usage

marcbachmann:
Just something small I found when reading through the code.
Thanks for this project ❤️


wrouesnel:
Thanks!

PR #211, opened 2018-08-22
FEATURE: support optional namespace caching

SamSaffron:
This allows us to execute slow queries like database size less frequently

Demo file is amended to demonstrate a 60 second cache for db size query


wrouesnel:
Part of me feels a metric proxy is a better candidate for this functionality, but I think merging this in if it helps people is just a good idea. Rebasing and merging.

SamSaffron:
Thanks for merging, I agree a metric proxy helps, the tricky thing though is that this provides extra flexibility, different metrics can have different cache intervals. In our particular case we only care to get db size updated once every 15 minutes cause it is expensive and the rest can have normal cadence. 

To achieve this with standard exporters without this feature we would need to run multiple exporters.

PR #212, opened 2018-08-23
Accept constantLabels as parameters

robsonpeixoto:
When you area monitoring server from cloud providers the exporter
IP/hostname is not the server IP, so add some constant labels is useful
on dashboards and alerts.

And is very important to use with `--disable-default-metrics` to add
info like `datname` and figure out the database name that you are
monitoring


wrouesnel:
This seems like a reasonable feature to add. Merged!

PR #215, opened 2018-09-04
multi server exporter is introduced

regeda:
Hi Everyone!
I've added an ability to define different data sources for one exporter setup.

Changes:
1. "server" label is added to each exported metric. "server" label is parsed from DSN and represents "host:port" pair.
2. To configure different data source you should list them in `DATA_SOURCE_NAME` separated by the comma.
3. I've added `disable-settings-metrics` flag to the command line in order to avoid panics in Prometheus client when different Postgres versions return different descriptions for settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,16 +49,19 @@ Package vendoring is handled with [`govendor`](https://github.com/kardianos/gove
   Path under which to expose metrics. Default is `/metrics`.
 
 * `disable-default-metrics`
-  Use only metrics supplied from `queries.yaml` via `--extend.query-path`
+  Use only metrics supplied from `queries.yaml` via `--extend.query-path`.
+
+* `disable-settings-metrics`
+  Use the flag if you don't want to scrape `pg_settings`.
 
 * `extend.query-path`
   Path to a YAML file containing custom queries to run. Check out [`queries.yaml`](queries.yaml)
   for examples of the format.
- 
+
 * `dumpmaps`
   Do not run - print the internal representation of the metric maps. Useful when debugging a custom
   queries file.
-  
+
 * `log.level`
   Set logging level: one of `debug`, `info`, `warn`, `error`, `fatal`
 
@@ -78,21 +81,23 @@ The following environment variables configure the exporter:
   URI may contain the username and password to connect with.
 
 * `DATA_SOURCE_URI`
-   an alternative to DATA_SOURCE_NAME which exclusively accepts the raw URI
+   an alternative to `DATA_SOURCE_NAME` which exclusively accepts the raw URI
    without a username and password component.
 
 * `DATA_SOURCE_USER`
   When using `DATA_SOURCE_URI`, this environment variable is used to specify
   the username.
+
 * `DATA_SOURCE_USER_FILE`
   The same, but reads the username from a file.
 
 * `DATA_SOURCE_PASS`
   When using `DATA_SOURCE_URI`, this environment variable is used to specify
   the password to connect with.
+
 * `DATA_SOURCE_PASS_FILE`
   The same as above but reads the password from a file.
-  
+
 * `PG_EXPORTER_WEB_LISTEN_ADDRESS`
   Address to listen on for web interface and telemetry. Default is `:9187`.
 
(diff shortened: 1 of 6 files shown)


wrouesnel:
This is good work. I'll need to give this a careful review - the need to disable the settings metrics makes me a bit wary of this type of change. There's some tension here against the idea that multi-server scraping should instead be driven by the Prometheus server and this exporter should act as a proxy for it.

wrouesnel:
@regeda Thanks very much for the PR! You are right, it's an in-demand feature and passing test-suites make me happy. Merging this PR to master now.

I'll hold off on creating a new explicit release till I get more of a chance to play with it, but it'll go out under the `wrouesnel/postgres_exporter:latest` image (which is the point of that tag for this repo anyway)

uspen:
Thank you for good work!

Can you add custom label or cluster_name from pg_settings for every metric by DSN?

regeda:
@uspen Pls, open the issue. I will support it.

karnatinaresh:
@regeda 
Can i define multiple endpoints with 
```
DATA_SOURCE_URI
DATA_SOURCE_USER
DATA_SOURCE_PASS
```
instead of `DATA_SOURCE_NAME`

regeda:
@karnatinaresh Unfortunately, multiple endpoints might be configured using `DATA_SOURCE_NAME` only

karnatinaresh:
@regeda Thanks

mlushpenko:
@regeda I tried multiple servers via URI and looks like it has problem with splitting a string:

```
time="2019-02-26T13:54:00Z" level=info msg="Established new database connection." source="postgres_exporter.go:1035"
time="2019-02-26T13:54:00Z" level=info msg="Error while closing non-pinging DB connection: <nil>" source="postgres_exporter.go:1041"
time="2019-02-26T13:54:00Z" level=info msg="Error opening connection to database (postgres://export:PASSWORD_REMOVED@localhost:5432/auth?sslmode=disable,postgres://wallets:test@localhost:5432/wallets?sslmode=disable): pq: unsupported sslmode \"disable,postgres://wallets:test@localhost:5432/wallets?sslmode=disable\"; only \"require\" (default), \"verify-full\", \"verify-ca\", and \"disable\" supported" source="postgres_exporter.go:1070"
time="2019-02-26T13:54:00Z" level=info msg="Starting Server: :9187" source="postgres_exporter.go:1178"
```

Also, doesn't hide password in second database string

regeda:
@mlushpenko Unfortunately, v0.4.7 didn't contain that feature. Use `master` branch instead.

mlushpenko:
Thanks you @regeda, no more problems with connection string. Now, with multiple databases I can't use simple queries, getting:
```
pq: relation \"users\" does not exist" source="postgres_exporter.go:1166"
pq: relation \"balances\" does not exist" source="postgres_exporter.go:1166"
```
It's probably a newbie question, but do you have idea how shall I reference tables with multiple connections? I tried:
`users` , `public.users` and `bp-auth.public.users` where `bp-auth` is database name

regeda:
The exporter establishes a separate connection for each data source. If a driver configured to connect with a database "XXX" then you can read relations existed in "XXX" database only.

mlushpenko:
HI @regeda separate connection sounds good. Could please clarify about driver? or give example of querying multiple databases with different tables from `queries.yml`? Maybe I can provide PGadmin screenshot as some help?

regeda:
@mlushpenko I'm not sure what do you want. A driver follows the database rules. You need to understand how Postgres works before diving into Go driver.

codestation:
I think he means how to specify the connection to use in queries.yml. For example if i define two queries but they are meant to different databases, how i restrict one query for one of the configured connections?

mlushpenko:
Thanks @codestation - exactly what I mean. Also, same thing here https://github.com/wrouesnel/postgres_exporter/issues/224#issuecomment-431790428. @regeda are we all missing something or can I give another example? 

If functionality for multiple databases was added, I don't understand why some example of queries weren't added. Maybe it's as simple as missing documentation or maybe I (and other people here) don't understand what was the purpose of separate connections if queries can work only for single connection.

regeda:
@mlushpenko As I understand you want to restrict queries by a specific database. Do not scrape all available schemas. Is it?

mlushpenko:
Yes, correct, thank you

karnatinaresh:
Hi,
I am using Vault to get db secrets for which i need a shell to expose these secrets as env variables but i unable to run a shell in this container.
Should I build a docker image with a different version of debian.

regeda:
@karnatinaresh I think you could build a docker image with any version of linux dist.

PR #223, opened 2018-10-08
Feature/disable settings metrics flag

itmecho:
This is just a simple flag that skips the pg_settings query. It was added to disable the exporter from exposing the settings found in the table.


itmecho:
I'm not quite sure what's causing the build to fail here, I've run `go run mage.go -v all` on my machine and it's completing successfully? I must be missing something?

wrouesnel:
This is now implemented via #215 .

PR #225, opened 2018-10-18
Exclude pg_settings_* on --disable-default-metrics

bitglue:
When using --disable-default-metrics, I expect postgres_exporter to
provide _no_ metrics besides those explicitly requested with
--extend.query-path.

See https://github.com/wrouesnel/postgres_exporter/issues/168#issuecomment-427337723


bitglue:
I just saw a similar thing has already implemented in #223. My preference would be to not introduce a new flag: the common use cases I can think of involve either wanting all the default metrics, or none of them. No reason to treat settings specially.

brosander:
We've set up an autobuild that builds v0.4.7 with this pr cherry picked on top:

[corvana/postgres-exporter:v0.4.7-no-pg-settings](https://hub.docker.com/r/corvana/postgres-exporter/tags/)

This is required for us to be able to scrape greenplum.

wrouesnel:
This feature is introduced by #215 is now merged to master.

PR #226, opened 2018-10-18
Document --constantLabels

bitglue:
This functionality was added in c225a1e1949e40d387b630683427cb0db259753f
but not documented.


bitglue:
I might also suggest renaming the flag to `--constant-labels` since none of the other flags use camelCase. Since the option isn't yet documented, maybe now is the best time?

wrouesnel:
And here's me cutting a release without looking at this specific PR...

I'll merge this, and guess I'll just support an alias for `--constant-labels` to `--constantLabels`

Issue #224, opened 2018-10-15
multiple databases in one  Instance


acalisku:
Hello,

Can exporter run for multiple databases in one postgres instance or is it just one database per exporter?

palmerabollo:
@acalisku It works with multiple databases in the same postgres server. Metrics are labelled with a `datname` label containing the database name they are associated to.

acalisku:
@palmerabollo  i mean , we need to add database specific metrics using queries.yaml , but because it just connects to one database , i won't be able to add those to exporter so , i think only solution for this for now  is to find a way to do cross database queries (maybe using fdw extension might help i guess) but it'd be nice to have an exporter that sees all databases but looks like it's not easy unless you can run multiple exporters

michalgutkowski:
Hi All, the change looks promissing... any idea when it can be released? 
The PR has comment about conflicts:
https://github.com/wrouesnel/postgres_exporter/pull/215

regeda:
Hi everyone.
I will resolve conflicts if the solution is reasonable. BTW, you can check out a fork https://github.com/gojuno/postgres_exporter

PR #230, opened 2018-11-08
Multi instance

gle4er:
Hello everyone. I just made fixes, to avoid issues like #224. Actually, this is rework of PR #215, because this PR didn't work for me.
In short: you should specify two or more URI's by comma (examples in Readme). You also can write usernames and passwords in files separated by comma or newlines.
My PR has tests, for multi instance check, so maintainers can support my fixes in future.
Also, I added flag for disabling pg_setting metrics (described in Readme too).


wrouesnel:
Hi, I just merged #215 which also introduced this feature. I'm closing this, but since it includes integration test changes expect to see some parts of it picked up before we create a release.

PR #237, opened 2018-11-20
added a flag to disable the collection of pg_settings as metrics

jbkc85:
looking for more ways to reduce the amount of metrics coming into the system, I was toying around with the `disableDefaultMetrics`, only to realize that the settings were still being collected.  Since I have no use for the settings, was wondering if it was at all possible to disable the collection of these settings through a similar argument.

btw, thanks for making work on this exporter...love it!


jbkc85:
yikes, saw that this was in another PR :-(

PR #244, opened 2018-12-20
fix-param-env-key fix spelling of env parameter

regeda:
fix of https://github.com/wrouesnel/postgres_exporter/issues/243


regeda:
@wrouesnel Could u merge the PR?

PR #247, opened 2019-01-24
Update non-superuser SQL commands in README

michaelkaye:
This makes the commands for running as non-superuser able to be re-run without as many errors.

This was tested as working on PG 9.4, so should be fairly stable.

This will help those who (like me) are automating these commands as part of a deployment, so we don't need to dive through postgresql docs. I don't think it will affect those running them only once.

I admit it doesn't simply fix the CREATE ROLE failing because the user already exists, but PG doesn't have a good way to do that as far as I could see. It's better than nothing anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,15 +176,14 @@ ALTER USER postgres_exporter SET SEARCH_PATH TO postgres_exporter,pg_catalog;
 -- If deploying as non-superuser (for example in AWS RDS), uncomment the GRANT
 -- line below and replace <MASTER_USER> with your root user.
 -- GRANT postgres_exporter TO <MASTER_USER>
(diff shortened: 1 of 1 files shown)


baburdick:
Failing `CREATE ROLE` can be handled in various situations by an aggregate of the answers here: https://stackoverflow.com/questions/8092086/create-postgresql-role-user-if-it-doesnt-exist

baburdick:
This should pull in the upstream changes to make it mergeable.

baburdick:
I agree this is better than nothing. But this should be closed in favor of #281, which is better still.

wrouesnel:
Closing in favor #281 which has been merged.

PR #266, opened 2019-04-16
Add option to automatically discover database names

arnogeurts:
This feature adds an option to dynamically retrieve the available databases / DSN's from a single DSN. This allows users to only configure a single DSN per database server and scrape metrics for all databases on the configured servers.


arnogeurts:
I feel like the failing build is not related to the changes made in this Pull Request, or am I missing something?

PR #281, opened 2019-06-15
Update README.md

asherf:
Allow non-superuser commands to be re-run without (some) errors.

Based on @michaelkaye PR https://github.com/wrouesnel/postgres_exporter/pull/247
I just resolved the merged conflicts on the original PR.
@baburdick FYI (I am happy to incorporate any feedback you may have as I am new to postgres)


baburdick:
@asherf: This looks like an improvement on #247.

Neither of these addresses the `CREATE ROLE` failure. But that's no reason not to merge this.

This appears to be the best answer on how to handle that without introducing a race condition: https://stackoverflow.com/a/55954480

PR #287, opened 2019-07-11
README: SQL that supports re-execution

sakishrist:
Make the SQL work in automated deployments where it will be repeated many times.

I needed to make the provided SQL part work in Ansible without elaborate checks on the Ansible part, so I modified some parts.

It might be overkill for the README file, so I leave that decision to the maintainer.

Cheers!


wrouesnel:
Idempotentency is always good.

PR #291, opened 2019-07-26
pg_up metric returns last calculated value without explicit resetting…

d-ulyanov:
If exporter is scraped by multiple Prometheuses (as we do) - Collect() could be called concurrently. In result in some cases one of Prometheuses could get pg_up = 0, because it was explicitly set to zero on first Collect call.

### DIFF
--- a/cmd/postgres_exporter/postgres_exporter.go
+++ b/cmd/postgres_exporter/postgres_exporter.go
@@ -128,6 +128,16 @@ type MetricMap struct {
 	conversion func(interface{}) (float64, bool) // Conversion function to turn PG result into float64
 }
 
+// ErrorConnectToServer is a connection to PgSQL server error
+type ErrorConnectToServer struct {
+	Msg string
+}
+
+// Error returns error
+func (e *ErrorConnectToServer) Error() string {
+	return e.Msg
+}
+
 // TODO: revisit this with the semver system
 func dumpMaps() {
 	// TODO: make this function part of the exporter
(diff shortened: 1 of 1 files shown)


cmd/postgres_exporter/postgres_exporter.go
```diff
@@ -812,21 +822,24 @@ func (s *Server) String() string {
...
-			log.Errorf("Error retrieving settings: %s", err)
-			errGauge.Inc()
+			err = fmt.Errorf("error retrieving settings: %s", err)
 		}
 	}
 
 	errMap := queryNamespaceMappings(ch, s)
 	if len(errMap) > 0 {
-		errGauge.Inc()
+		err = fmt.Errorf("queryNamespaceMappings returned %d errors")
```
vlamug: fmt.Errorf("queryNamespaceMappings returned %d errors", len(errMap))

cmd/postgres_exporter/postgres_exporter.go
```diff
@@ -1280,16 +1293,42 @@ func (e *Exporter) scrape(ch chan<- prometheus.Metric) {
...
 	}
+
+	var errorsCount int
+	var hasConnectionErrors bool
+
 	for _, dsn := range dsns {
-		e.scrapeDSN(ch, dsn)
+		err := e.scrapeDSN(ch, dsn)
+
+		if err != nil {
```
vlamug: `if err := e.scrapeDSN(ch, dsn); err != nil {`

cmd/postgres_exporter/postgres_exporter.go
```diff
@@ -1330,24 +1369,18 @@ func (e *Exporter) discoverDatabaseDSNs() []string {
...
 }
 
-func (e *Exporter) scrapeDSN(ch chan<- prometheus.Metric, dsn string) {
+func (e *Exporter) scrapeDSN(ch chan<- prometheus.Metric, dsn string) error {
 	server, err := e.servers.GetServer(dsn)
 	if err != nil {
-		log.Errorf("Error opening connection to database (%s): %v", loggableDSN(dsn), err)
-		e.error.Inc()
-		return
+		return &ErrorConnectToServer{fmt.Sprintf("Error opening connection to database (%s): %v", loggableDSN(dsn), err)}
```
vlamug: I think it is better to use %s placeholder for error

cmd/postgres_exporter/postgres_exporter.go
d-ulyanov: thx

cmd/postgres_exporter/postgres_exporter.go
d-ulyanov: okay)

d-ulyanov:
@wrouesnel pls take a look for this issue

PR #299, opened 2019-08-12
Yaml parsing refacto

AlexisSellier:
Use `struct` instead of `interface{}` while parsing user queries file.
I split the function so it easier to test it.

I also added a basic test to start strengthening the code

I did not use the unmarshalled struct in the rest of the code just to be sure I do not break something. Will do it in a near future


nouney:
Please merge

Issue #231, opened 2018-11-08
SCRAM support


keithf4:
Appears that postgres_exporter doesn't support when SCRAM password authentication is set in pg_hba.conf. 

https://www.postgresql.org/docs/10/auth-methods.html#AUTH-PASSWORD

```
Nov  8 10:52:30 localhost postgres_exporter: time="2018-11-08T10:52:30-05:00" level=info msg="Error opening connection to database (postgresql://ccp_monitoring:PASSWORD_REMOVED@localhost:5432/postgres?sslmode=disable): pq: unknown authentication response: 10" source="postgres_exporter.go:1030"
```
Thankfully just setting an override line above other entries in pg_hba.conf for the postgres_exporter role to md5 allows it to work for now even when the internal password storage is scram.

ptman:
The support is needed in lib/pq: https://github.com/lib/pq/pull/608 or https://github.com/lib/pq/pull/788

onderaycicek:
This issue solved. https://github.com/lib/pq/pull/833

Neustradamus:
Linked to:
- https://github.com/lib/pq/issues/817
- https://github.com/lib/pq/pull/608
- https://github.com/lib/pq/pull/621
- https://github.com/lib/pq/pull/788
- https://github.com/lib/pq/pull/833

jesperpedersen:
https://github.com/lib/pq/releases/tag/v1.1.1 contains the support for this

RobAtticus:
Wanted to resurface this @wrouesnel because we have found it to be quite important in a project we're working on. I see there's a PR (#278) to address this, although it's biting off a lot in addition. Is it possible to get this change in by itself, which seems to be just updating the pq dependency to 1.1.1?

jesperpedersen:
Yeah, and https://github.com/lib/pq/releases/tag/v1.2.0 is out now

wrouesnel:
Resolved in v0.6.0 and latest.

brunolnetto:
Hi, folks. I start postgres_exporter.service on Linux 20.04. I am unable to connect to the target database. Hhe resuling log is `pq: unknown authentication response: 10`. It seems, there is some issue with `SCRAM-SHA-256` I am not aware of. May you provide the steps or commands necessary to resolvethis issue?

PR #304, opened 2019-08-22
Support SCRAM-SHA-256

wlfshmn:
Updated lib/pg driver to 1.2.0 in order to support stronger SCRAM-SHA-256 authentication. This drops support for Go < 1.11 and PostgreSQL < 9.4.


wlfshmn:
Intended as a minimal fix for #231 which was blocking us from monitoring new servers.

RobAtticus:
Would love to see this in as soon as possible -- thanks @wlfshmn

iand:
We're currently maintaining an internal fork with just this commit added. It would be great to get this updated since the version of lib/pq currently bundled is two years old @wrouesnel

RobAtticus:
Hi @wrouesnel, any idea if this can be merged soon?

wrouesnel:
Merged.

Neustradamus:
@wlfshmn and others: Thanks a lot for all about SCRAM.

Linked to:
- https://github.com/scram-xmpp/info/issues/1

PR #307, opened 2019-09-17
Expose pg_current_wal_lsn_bytes

michaelkaye:
pg_current_wal_lsn is currently DISCARDed here: https://github.com/wrouesnel/postgres_exporter/blob/master/cmd/postgres_exporter/postgres_exporter.go#L244

I believe this is because it's currently a string value that is the position within the WAL logs, and isn't immediately useful in graphing.

This PR adds a metric `pg_current_wal_lsn_bytes`, which is the diff between the current position and the initial WAL position of `0/0` -- effectively a translation of the WAL log position into bytes, such that `rate(pg_stat_replication_pg_currentl_wal_lsn_bytes)` can be meaningful on graphs.

We use this metric to see the rate of WAL entries being created on the primary, independently of whether the primary and secondary are becoming further behind or catching up - you can then tell the difference between the WAL creating a lot more entries than normal, but replication is proceeding at the normal rate; and the WAL having the same number of entries, but replication is proceeding slower than normal.

I've considered making this a local change in a queries.yml file, which we do use for other things - `pg_stat_user_tables`, for instance, however it feels odd to have just this one `pg_stat_replication` related metric specified in there and the rest in the main codebase.

Could also consider exposing this directly as `pg_current_wal_lsn`, as we've been discarding it so long.


michaelkaye:
lag graph

(for instance here is when we used the metric to see that the replication is not due to an increase in the amount going in the WAL, and must be an issue with the remote server)

PR #324, opened 2019-11-01
Export pg_stat_archiver metrics

russss:
These are useful for knowing if WAL archiving is working successfully.

I believe the pg_stat_archiver view was added in 9.4 (which is the oldest currently-supported version of postgres), but it's not clear whether I can stop this being queried in older versions.

### DIFF
--- a/cmd/postgres_exporter/postgres_exporter.go
+++ b/cmd/postgres_exporter/postgres_exporter.go
@@ -278,6 +278,19 @@ var builtinMetricMaps = map[string]intermediateMetricMap{
 		},
 		0,
 	},
+	"pg_stat_archiver": {
+		map[string]ColumnMapping{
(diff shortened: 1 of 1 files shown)


wrouesnel:
I'm happy with the test suite passing the integration tests, which represent the sort of limits of what I want to guarantee. This looks good so merging.

nikhil-bhat:
@russss @wrouesnel  I am getting repeated errors because version of posgres(9.2.23) doesnt have this view . COuld you let me know how i can disable this as this causes postgres logs to fill up 
2019-12-12 08:05:43 EST:::1(52124):postgres@postgres:[181027]:ERROR:  relation "pg_stat_archiver" does not exist at character 94

nikhil-bhat:
I am willing to create a PR for this

PR #333, opened 2019-11-06
Build in Docker

russss:
I develop on a Mac, and when I tried to build a docker image using the method in the README it naturally failed because it was trying to run a darwin binary on the linux docker image.

This PR runs the Go build within Docker, which means I can build on my Mac. It also means that the version of Go used is consistent across builds as it's specified in the Dockerfile.

I haven't tried to change the CI build, so this PR is not complete, but I thought I'd submit it in case it was of interest.


dannietjoh:
you could add a Dockerfile.build as a separate file

rhysjtevans:
Why wouldn't this be the default approach, definitely should be merged.

SuperQ:
We're now using the Prometheus build system.

PR #386, opened 2020-03-24
ARM64 platform

v12aml:
worked fine on arm64


muhlba91:
Would you mind drafting a release including this change to have a release including binaries for arm64?

PR #398, opened 2020-05-22
Add pg_stat_statements view for non-superuser instructions

robbiet480:
Needed this to fix errors like `* collected metric "pg_stat_statements_blk_write_time_seconds" { label:<name:"datname" value:"postgres" > label:<name:"rolname" value:"postgres_exporter" > label:<name:"server" value:"localhost:5432" > counter:<value:0 > } was collected before with the same name and label values`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,18 @@ AS
   SELECT * FROM get_pg_stat_replication();
 
 GRANT SELECT ON postgres_exporter.pg_stat_replication TO postgres_exporter;
+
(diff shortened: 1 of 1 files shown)


jkroepke:
👍 Works for me.

anarcat:
it should be noted that this doesn't work unless `pg_stat_statements` is loaded, which is not the case by default on Debian buster 10 (psql 11).

jkroepke:
@anarcat I'm not a maintainer here, but you can just raise a pull request like this to add the note.

PR #412, opened 2020-07-08
Get replication lag information from pg_stat_replication

azlev:
Convert write_lag, flush_lag and replay_lag from interval to float before expose the information.


wrouesnel:
Changing built in collectors to depend on a Postgres database function isn't a call for the exporter to make - this is the type of thing which should be supplied via a queries.yml file.

PR #414, opened 2020-07-20
Recover when connection cannot be established straight at startup

yann-soubeyrand:
When the connection to the PostgreSQL instance cannot be established straight
at startup, a race condition can happen when autoDiscoverDatabases is true. If
discoverDatabaseDSNs fails, no dsn is set as the master database, and, if
scrapeDSN succeeds, checkMapVersions will have omitted the default metrics in
the server metric map. The metric map won't be updated unless the version
returned by the PostgreSQL instance changes. With this patch, scrapeDSN won't
be run unless discoverDatabaseDSNs succeeded and thus the race condition is
eliminated.

Signed-off-by: Yann Soubeyrand <yann.soubeyrand@camptocamp.com>

### DIFF
--- a/cmd/postgres_exporter/postgres_exporter.go
+++ b/cmd/postgres_exporter/postgres_exporter.go
@@ -1503,12 +1503,12 @@ func (e *Exporter) discoverDatabaseDSNs() []string {
 			continue
 		}
 
-		dsns[dsn] = struct{}{}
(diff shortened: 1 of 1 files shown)


yann-soubeyrand:
Closing in favor of #415 (I used master branch instead of a feature branch and didn't manage to change the PR's branch).

PR #428, opened 2020-09-03
User defined configuration to run the query for different database server version from single yml file

neel5481:
User can define the range of database server version where query is to be executed in yml configuration file.

If user want to run the query on database version greater than 10.0.0, use below format.
      runonserver: ">=10.0.0"

      Below are the example of db version range user can defined in yml file.
            <=10.1.0
            >=12.1.0
            =11.0.0
            <9.6.0 || >=11.0.0

### DIFF
--- a/cmd/postgres_exporter/postgres_exporter.go
+++ b/cmd/postgres_exporter/postgres_exporter.go
@@ -1456,7 +1456,7 @@ func queryNamespaceMappings(ch chan<- prometheus.Metric, server *Server) map[str
 			serVersion, _ := semver.Parse(server.lastMapVersion.String())
 			runServerRange, _ := semver.ParseRange(server.runonserver)
 			if !runServerRange(serVersion) {
-				log.Debugln("Query skipped for database version: [%s] as it should be run on database server version: [%s]", server.lastMapVersion.String(), server.runonserver)
+				log.Debugln("Query skipped for database version: ", server.lastMapVersion.String(), " as it should be run on database server version: ", server.runonserver)
 				continue
 			}
 		}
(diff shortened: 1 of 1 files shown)


neel5481:
@wrouesnel Can you please update what is wrong here ?

cobolbaby:
@neel5481  Have a few questions:

- How to load the configuration? `--extend.query-path`?
- Server.runonserver have nothing to do with metric cycle acquisition
- Should the UserQuery.RunOnServer be used in `parseUserQueries` function?

neel5481:
@cobolbaby When user want to execute queries on specific database only to collect metrics, "RunOnServer" will be useful and it will be used along with "PG_EXPORTER_EXTEND_QUERY_PATH" where user defined the queries in yaml file and give to the exporter. You can check detail usage in the commit message.

cobolbaby:
@neel5481 After trying and debugging, I found that only the exporter's startup configuration `runonserver` does work, but the `runonserver` in the yml file actually had no practical effect.

cobolbaby:
https://github.com/prometheus-community/postgres_exporter/blob/1dc210476ebd630aec7feff4c4487163724aedd3/cmd/postgres_exporter/postgres_exporter.go#L1468-L1476

Are you sure that the above logic should be placed in the loop logic?

neel5481:
> https://github.com/prometheus-community/postgres_exporter/blob/1dc210476ebd630aec7feff4c4487163724aedd3/cmd/postgres_exporter/postgres_exporter.go#L1468-L1476
> 
> Are you sure that the above logic should be placed in the loop logic?

Yes, you are right. Do you have any suggestion where should we add this skip block after reading the actual database version and runonserver given in yaml file, before I debug ?

neel5481:
> https://github.com/prometheus-community/postgres_exporter/blob/1dc210476ebd630aec7feff4c4487163724aedd3/cmd/postgres_exporter/postgres_exporter.go#L1468-L1476
> 
> Are you sure that the above logic should be placed in the loop logic?

Can you please verify this pull request ?

https://github.com/prometheus-community/postgres_exporter/pull/509

cobolbaby:
Ok

PR #434, opened 2020-09-18
Introduce histogram support

### LINKED ISSUE #402: Histogram queries are not supported
queries.yaml does not include any "histogram" metrics.

Can support for "histogram" metrics be added?

au-phiware:
Prior to this change, the custom queries were restricted to counters and gauges.

This change introduces a new ColumnUsage, namely HISTOGRAM, that expects the column to contain an array of upper inclusive bounds for each observation bucket in the emitted metric.  It also expects three more columns to be present with the suffixes:
- `_bucket`, containing an array of cumulative counters for the
  observation buckets;
- `_sum`, the total sum of all observed values; and
- `_count`, the count of events that have been observed.

A flag has been added to the MetricMap struct to easily identify metrics that should emit a histogram and the construction of a histogram metric is aided by the pg.Array function and a new helper dbToUint64 function.

Finally, and example of usage is given in queries.yaml.

fixes #402

This was something I needed on a last_seen timestamp type of column in my database.


au-phiware:
Moved to a different branch, apologies for the confusion.

PR #453, opened 2020-10-21
aws rds iam auth

kppullin:
Add support for AWS RDS IAM authentication


ArthurSens:
Thank you for the significant effort that went into adding RDS IAM authentication and working through the exporter integration. The exporter and its dependency setup have changed considerably since 2020, so this branch no longer applies cleanly and the large vendored dependency snapshot would need to be revisited. We are closing it as part of the backlog cleanup, while keeping the implementation here as useful prior art. If this support is pursued again, we would welcome a fresh implementation against the current codebase.

PR #463, opened 2020-12-18
change windows build binary error

killerwzb:
Signed-off-by: killerwzb <killerwzb@163.com>


SuperQ:
We've replaced mage with promu.

Issue #293, opened 2019-07-31
`--auto-discover-databases` is failing DSN validation


fahrradflucht:
Following https://github.com/wrouesnel/postgres_exporter#setting-the-postgres-servers-data-source-name and https://github.com/wrouesnel/postgres_exporter#automatically-discover-databases I expected this to work:
```
bash-5.0$ DATA_SOURCE_NAME="user=postgres host=/var/run/postgresql/ sslmode=disable" postgres_exporter --auto-discover-databases --log.level debug
INFO[0000] Established new database connection to "/var/run/postgresql/:5432".  source="postgres_exporter.go:777"
DEBU[0000] Querying Postgres Version on "/var/run/postgresql/:5432"  source="postgres_exporter.go:1212"
INFO[0000] Semantic Version Changed on "/var/run/postgresql/:5432": 0.0.0 -> 11.4.0  source="postgres_exporter.go:1229"
DEBU[0000] pg_xlog_location_diff is being forced to discard due to version incompatibility.  source="postgres_exporter.go:504"
DEBU[0000] replay_location is being forced to discard due to version incompatibility.  source="postgres_exporter.go:504"
DEBU[0000] flush_location is being forced to discard due to version incompatibility.  source="postgres_exporter.go:504"
DEBU[0000] write_location is being forced to discard due to version incompatibility.  source="postgres_exporter.go:504"
DEBU[0000] sent_location is being forced to discard due to version incompatibility.  source="postgres_exporter.go:504"
DEBU[0000] procpid is being forced to discard due to version incompatibility.  source="postgres_exporter.go:504"
DEBU[0000] Querying pg_setting view on "/var/run/postgresql/:5432"  source="pg_setting.go:15"
DEBU[0000] Querying namespace:  pg_stat_activity         source="postgres_exporter.go:1192"
DEBU[0000] Querying namespace:  pg_stat_bgwriter         source="postgres_exporter.go:1192"
DEBU[0000] Querying namespace:  pg_stat_database         source="postgres_exporter.go:1192"
DEBU[0000] Querying namespace:  pg_stat_database_conflicts  source="postgres_exporter.go:1192"
DEBU[0000] Querying namespace:  pg_locks                 source="postgres_exporter.go:1192"
DEBU[0000] Querying namespace:  pg_stat_replication      source="postgres_exporter.go:1192"
ERRO[0000] Error opening connection to database (postgres): malformed dsn "postgres"  source="postgres_exporter.go:1336"
[...]

PR #473, opened 2021-01-23
Support connstring syntax when discovering databases

intelfx:
Support connstring DSNs (`host=... user=... password=... dbname=...`) in
addition to URIs (`postgresql://user:pass@host/dbname`) for purposes of
database discovery.

Connstring syntax is needed to support accessing PostgreSQL via Unix
domain sockets (`host=/run/postgres`), which is not really possible with
URI syntax.

Fixes: #293


intelfx:
OK, fixed an unrelated CI failure while at it.

PR #509, opened 2021-03-30
Exporter's startup configuration 'runonserver' does work but the 'runonserver' from yml file has not effect.

neel5481:
Exporter's startup configuration 'runonserver' does work but the 'runonserver' from yml file has not effect.

Reported by: cobolbaby

### DIFF
--- a/cmd/postgres_exporter/postgres_exporter.go
+++ b/cmd/postgres_exporter/postgres_exporter.go
@@ -517,7 +517,7 @@ func makeQueryOverrideMap(pgVersion semver.Version, queryOverrides map[string][]
 	return resultMap
 }
 
-func parseUserQueries(content []byte) (map[string]intermediateMetricMap, map[string]string, error) {
+func parseUserQueries(content []byte, server *Server) (map[string]intermediateMetricMap, map[string]string, error) {
 	var userQueries UserQueries
 
 	err := yaml.Unmarshal(content, &userQueries)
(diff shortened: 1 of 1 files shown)


cmd/postgres_exporter/postgres_exporter.go
```diff
@@ -517,7 +517,7 @@ func makeQueryOverrideMap(pgVersion semver.Version, queryOverrides map[string][]
 	return resultMap
 }
 
-func parseUserQueries(content []byte) (map[string]intermediateMetricMap, map[string]string, error) {
+func parseUserQueries(content []byte, server *Server) (map[string]intermediateMetricMap, map[string]string, error) {
```
cobolbaby, line 520: `parseUserQueries` should compare the current server version with the version of the custom metric configuration.

It's strange that you make logical judgments in this loop.

cmd/postgres_exporter/postgres_exporter.go
```diff
@@ -531,6 +531,17 @@ func parseUserQueries(content []byte) (map[string]intermediateMetricMap, map[str
 
 	for metric, specs := range userQueries {
 		level.Debug(logger).Log("msg", "New user metric namespace from YAML metric", "metric", metric, "cache_seconds", specs.CacheSeconds)
+
+                // check if the query is to be run on specific database server version range or not
+                if len(specs.RunOnServer) > 0 {
+                        serVersion, _ := semver.Parse(server.lastMapVersion.String())
+                        runServerRange, _ := semver.ParseRange(specs.RunOnServer)
```
ArthurSens, line 538:
```suggestion
                        runServerRange, err := semver.ParseRange(specs.RunOnServer)
```

we should not ignore err here.

cmd/postgres_exporter/postgres_exporter.go
```diff
@@ -571,7 +582,7 @@ func parseUserQueries(content []byte) (map[string]intermediateMetricMap, map[str
 // TODO: test code for all cu.
 // TODO: the YAML this supports is "non-standard" - we should move away from it.
 func addQueries(content []byte, pgVersion semver.Version, server *Server) error {
-	metricMaps, newQueryOverrides, err := parseUserQueries(content)
+	metricMaps, newQueryOverrides, err := parseUserQueries(content, server)
```
ArthurSens, line 585: why pass server if pgVersion is already an argument?

ArthurSens (changes requested):
Hi 👋 , sorry for the delayed review here! Even though the PR is 5 years old, it seems like the bug still exists 😅 

We still need to make a few improvements to make this PR mergeable, starting with a rebase. Second, the current code doesn't compile 😛.

A good practice is also to first start with a test that identifies the bug, and then we can add extra commits that fix the test.

PR #524, opened 2021-04-26
Small improvements on Documentation, more secure way to run the exporter

josevnz:
Hello,

Thanks a lot for writing the exporter, it is a very useful piece of software!. 

I made a few changes on the README.md file in regard how you can setup the datasource using an environment file, which is more secure than passing an --env. Also it is best practice to use long options, specially with docker, so the intent is is clear.

If you think this is useful it would be great to have this merged into the trunk of your code.

Thanks,

--Jose Vicente Nunez.


josevnz:
It using an inactive account, closing this pull request.

Issue #502, opened 2021-03-14
View pg_stat_statements broken in pg13


parisni:
several fields have been renamed in pg 13 (total_time -> total_exec_time). Then the below query is changed to make the queries.yaml working

```sql
  query: "SELECT t2.rolname, t3.datname, queryid, calls, total_exec_time / 1000 as total_time_seconds, min_exec_time / 1000 as min_time_seconds, max_exec_time / 1000 as max_time_seconds, mean_exec_time / 1000 as mean_time_seconds, stddev_exec_time / 1000 as stddev_time_seconds, rows, shared_blks_hit, shared_blks_read, shared_blks_dirtied, shared_blks_written, local_blks_hit, local_blks_read, loc
al_blks_dirtied, local_blks_written, temp_blks_read, temp_blks_written, blk_read_time / 1000 as blk_read_time_seconds, blk_write_time / 1000 as blk_write_time_seconds FROM pg_stat_statements t1 JOIN 
pg_roles t2 ON (t1.userid=t2.oid) JOIN pg_database t3 ON (t1.dbid=t3.oid) WHERE t2.rolname != 'rdsadmin'"```

SuperQ:
Want to open a PR to update it?

parisni:
not sure, because the query won't work on pg < 13 . A PR means introduce a dedicated yaml file ?

cobolbaby:
You can try to config the `runonserver` in *.yml.

https://github.com/prometheus-community/postgres_exporter/blob/c51a28f5f086f4f011ad9b4c9dcd4e78d1cb2c12/cmd/postgres_exporter/postgres_exporter.go#L121-L127

mr-tron:
Does anybody know how it can be fixed without kludges via custom queries.yml?

vl4deee11:
> Does anybody know how it can be fixed without kludges via custom queries.yml?

Added fix for version >= 13.0.0

veera-chenna:
Facing the same issue. This [PR-553](https://github.com/prometheus-community/postgres_exporter/pull/553) is awsome. Can someone review the PR and release it as a new feature?

rasta-rocket:
Yep same here ^^

Timusan:
Running into this as well. PG 14 has been released for a while now, this exporter breaks there to. Someone might want to review PR-553?

combrs:
Also encounter this. 
I see some conflict in warning about deprecate custom query.yml 
[...]

PR #543, opened 2021-06-10
Versions in queries.yaml

vl4deee11:
Fix issue #502 

Added template for breaking changes

This will help users redefine the fields for their specific version without editing the project code.


vl4deee11:
@SuperQ Check this PR please

vl4deee11:
@SuperQ please check this PR =^)

vl4deee11:
@SuperQ  can you react in any way

SuperQ:
It's an interesting idea. I'm not sure I like how it works.

vl4deee11:
@SuperQ Okay, don't you like it in terms of code design or in terms of performance ? In fact, it will help people to keep in the same values.yaml settings for databases of different versions

SuperQ:
I think if we need to get more complicated than what can be done with queries.yaml, we should move the query into a dedicated collector.

vl4deee11:
@SuperQ  okay, I'll try to do it

SuperQ:
Thanks! I would love to see more of the queries.yaml functionality moved into code. This makes it easier for users. (This is how the MySQL exporter works)

vl4deee11:
I looked at all the code as much as I could during my work time and also looked at mysql_exporter a little, and I did not understand a bit what you want to see with the versions, could you help a little with this?

SuperQ:
The whole exporter needs a bit of a refactor to make it easier to add features. I know @sysadmind was interested in doing some work on this.

vl4deee11:
Yes, I agree that this requires refactoring, but unfortunately, I cannot take responsibility for refactoring, as I am afraid that I will not have enough time (now loaded with university studies and work), but at the same time I could help with this

sysadmind:
I did start trying to clean up the code to make a refactor easier. I opened #552 as a start. I think making sure that we have the appropriate test coverage will make sure we don't lose behavior unintentionally. I'll take a look at this pr when I get some time.

vl4deee11:
@sysadmind Ok, thanks. I'll be happy to inject queries with versions into the new code (which I wanted to do in this PR), if you need help with refactoring, I could help (write to vl4deee11@gmail.com)!

vl4deee11:
New https://github.com/prometheus-community/postgres_exporter/pull/553

PR #556, opened 2021-07-07
refactor pg_stat_bgwriter metrics into standalone collector

sysadmind:
This moves the metrics that are queried from pg_stat_bgwriter into a dedicated collector instead of dynamically generating queries and metrics from a map. It adopts a new, consistent namespace `postgres` for its metrics and includes the `_total` suffix on all of the counters to match prometheus standards. 

This implementation uses QueryRowContext to allow for later addition of context for cancellation. The Postgres docs state that this should only ever be a single row.

This ignores any sort of future Collector interface to keep the surface area small and to allow for that interface to get fleshed out before implementation.

The previous implementation had a label `server`, so I kept it but I am unsure if it's even useful. 

@SuperQ This should not be merged before the v0.10.0 tag. I am also very open to feedback as this is the first step of many in the refactor.

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -87,10 +87,6 @@ type PostgresCollector struct {
 	logger     log.Logger
 
 	servers map[string]*server
-
-	// autoDiscoverDatabases will cause the collector to query the database
-	// to find other servers and also scrape them.
-	autoDiscoverDatabases bool
 }
 
 type Option func(*PostgresCollector) error
@@ -146,12 +142,6 @@ func NewPostgresCollector(logger log.Logger, dsns []string, filters []string, op
 		if err != nil {
 			return nil, err
 		}
-		// Manually provided servers are always classified as "primary"
-		s.isPrimary = true
-
-		// TODO(@sysadmind): We need to discover the downstream servers and add them here.
-		// if p.autoDiscoverDatabases {
-		// }
 
 		servers[dsn] = s
 	}
@@ -159,13 +149,6 @@ func NewPostgresCollector(logger log.Logger, dsns []string, filters []string, op
 	p.servers = servers
 
 	return p, nil
-}
-
-func WithAutoDiscoverDatabases(discover bool) Option {
-	return func(p *PostgresCollector) error {
-		p.autoDiscoverDatabases = discover
-		return nil
-	}
 }
 
 // Describe implements the prometheus.Collector interface.
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -87,10 +87,27 @@ type PostgresCollector struct {
 	logger     log.Logger
 
 	servers map[string]*server
+
+	// autoDiscoverDatabases will cause the collector to query the database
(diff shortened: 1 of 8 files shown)


sysadmind:
@SuperQ This should be ready to review now

SuperQ:
Thanks, sorry for the delay in reviewing.

I have some concerns about renaming the namespace on these metrics. But It may be a good idea long-term. I guess we need to make that decision first.

sysadmind:
@SuperQ Is that something to discuss here or an issue or would that be better posed to the mailing list?

sysadmind:
@SuperQ I refactored this now with the new collector registration and reverted the namespace changes. We can always revisit this in the future.

collector/pg_stat_bgwriter.go
```diff
@@ -0,0 +1,212 @@
...
+	return &PGStatBGWriterCollector{}, nil
+}
+
+const bgWriterSubsystem = "stat_bgwriter"
+
+var statBGWriter = map[string]*prometheus.Desc{
+	"checkpoints_timed": prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, bgWriterSubsystem, "checkpoints_timed_total"),
+		"Number of scheduled checkpoints that have been performed",
+		[]string{"server"},
```
SuperQ, line 41: I just realized I think we have something going wrong with these labels. This seems like it's a discovery label. Each scrape should only ever have one `server` in the output.

collector/collector.go
```diff
@@ -121,48 +138,62 @@ func NewPostgresCollector(logger log.Logger, dsns []string, filters ...string) (
...
 		s, err := makeServer(dsn)
 		if err != nil {
 			return nil, err
 		}
+		// Manually provided servers are always classified as "primary"
+		s.isPrimary = true
+
+		// TODO(@sysadmind): We need to discover the downstream servers and add them here.
+		// if p.autoDiscoverDatabases {
+		// }
```
SuperQ, line 154: I think we're intending to drop the `autoDiscoverDatabases` feature. IIRC It's misused by the queries.yaml to turn this exporter into a generic SQL exporter.

sysadmind: I'm happy to hear that. I was going to ask you about the usage of this flag because it is challenging to understand from the code. I'm not surprised that it causes issues for users.

SuperQ: Yea, it's a confusing mess. Part of why I haven't done much maintenance work on this exporter myself. :joy_cat:

collector/pg_stat_bgwriter.go
sysadmind: The DSNs passed in can be multiple so this would be the only way to distinguish metrics between them.

SuperQ: Ugh, this exporter is so broken. :-(

SuperQ (approved):
LGTM, Thanks!

PR #558, opened 2021-07-13
feat: scrape timeout

johanneswuerbach:
Allow to configure a scrape timeout to prevent queries from hanging forever.

### DIFF
--- a/cmd/postgres_exporter/datasource.go
+++ b/cmd/postgres_exporter/datasource.go
@@ -14,6 +14,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
 	"net/url"
@@ -25,7 +26,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-func (e *Exporter) discoverDatabaseDSNs() []string {
+func (e *Exporter) discoverDatabaseDSNs(ctx context.Context) []string {
 	// connstring syntax is complex (and not sure if even regular).
 	// we don't need to parse it, so just superficially validate that it starts
 	// with a valid-ish keyword pair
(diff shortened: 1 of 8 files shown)


roidelapluie:
Prometheus sends the scrape timeout in HTTP headers, we should reuse it.

johanneswuerbach:
While I've seen that, I wasn't sure how well this plays with caching of slow queries 🤔

johanneswuerbach:
Looking at https://github.com/prometheus/client_golang support for the `X-Prometheus-Scrape-Timeout-Seconds` header doesn't seem to be provided yet, any pointers how that works @roidelapluie , @SuperQ 

This exporter is using the handler provided by the golang client at the moment https://github.com/prometheus-community/postgres_exporter/blob/57719ba53cac428769aaf3c4c0bb742df3cfca98/cmd/postgres_exporter/main.go#L117

sysadmind:
I do not think that the context is propagated to any of the collectors at this time. While the http.Request may be cancelled by the header timeout, the downstream functions in the collector have no way to access it. I think this implementation is _reasonable_ at this time based on the interfaces that we have to work with.

johanneswuerbach:
Any further steps I could do to drive this forward @sysadmind @roidelapluie @SuperQ ?

johanneswuerbach:
Ping :-)

SuperQ:
Yes, in order to do this the "usual way", you need to implement `http.HandlerFunc()`.

The [mysqld_exporter does this](https://github.com/prometheus/mysqld_exporter/blob/3d2a10b014e2268890d0aff57de319b6ff78ddd7/mysqld_exporter.go#L184-L239).

I think we should focus on getting https://github.com/prometheus-community/postgres_exporter/pull/618 finished so we have the new multi-target handler functionality.

ArthurSens:
Hi 👋, I'm closing this PR because it hasn't been active for a while, and the codebase has changed a lot over the last few years. #618 has landed, but it still doesn't solve the long-running queries without cancellation problem.

I think we need to advance https://github.com/prometheus/client_golang/issues/1538, and once that lands, we can bring those improvements here.

Issue #549, opened 2021-06-29
`pg_stat_statement` metrics have unreasonable cardinality


lichuan0620:
**What did you do?**

deploy postgres-exporter, run lots of queries

**What did you expect to see?**

everything works, nothing crashes

**What did you see instead? Under which circumstances?**

The number of time series created by postgres-exporter increased rapidly. Prometheus was OOM killed soon after.

**Additional comment**

I understand that it's commonly agreed that Prometheus metrics should have reasonably cardinality and avoid ID-like labels such as trace ID or query ID. This best practice has been discussed in [this post](https://www.robustperception.io/cardinality-is-key) and various community issues. 

Of course, the user could always disable or drop these metrics, as I already have. But these are still relevant information and could cause confusion to those who haven't investigate closely (It did in my team). These data should be organized into more well reasonable metrics that either sum over different queries or put them into histograms.

**Environment**

* System information:

	insert output of `uname -srm` here

* postgres_exporter version:

	0.8.0

* postgres_exporter flags:

```
insert list of flags used here
```

* PostgresSQL version:

	insert PostgreSQL version here

* Logs:
```
insert logs relevant to the issue here
```

Harkishen-Singh:
@lichuan0620 does the cardinality issue occurs due to your too many custom queries, or the default query?

sysadmind:
I don't think that pg_stat_statements is part of the default metrics exported by the exporter. Did you copy the example queries.yaml file and set it to be used with `--extend.query-path` or `PG_EXPORTER_EXTEND_QUERY_PATH`? You could remove it from that file to not export those metrics. It's probably a poor metric to export and we should remove it from the example.

lichuan0620:
> You could remove it from that file to not export those metrics.

Yes, and I've done just that. 

> we should remove it from the example

That'll do.

SuperQ:
I made a [filtered version of the `pg_stat_statments`](https://gitlab.com/gitlab-cookbooks/gitlab-exporters/-/blob/b33a2d7090b668205e94616df7c793a472d0d88e/templates/postgres_exporter/queries.yaml.erb#L282-328) query a while back. This reduced our cardinality a lot, and put some harder limits on it.

```yaml
pg_stat_statements:
  query: |
    SELECT
      pg_get_userbyid(userid) as user,
      pg_database.datname,
      pg_stat_statements.queryid,
      pg_stat_statements.calls,
[...]

PR #561, opened 2021-07-16
Add a warning for pg_stat_statements

sysadmind:
This metric can be very expensive to store in prometheus and has the potential to create an unreasonably large cardinality set of metrics so warn the user.

This was brought up in #549 

ping @SuperQ 


SuperQ (approved):
LGTM

PR #590, opened 2021-11-17
#258 Document running as non-superuser on PostgreSQL versions 10 or newer.

### LINKED ISSUE #258: Suggest using `pg_monitor` role for PostgreSQL >= 10
As of PostgreSQL 10, there are [built-in roles](https://www.postgresql.org/docs/10/default-roles.html) that give access to extra statistics without needing to be superuser.

The documentation should suggest using creating an unprivileged user and granting the `pg_monitor` role to it.

juris-greitans:
Starting from version 10, PostgreSQL includes a built-in role `pg_monitor` meant for use by monitoring tools. This PR adds documentation about granting the role when running postgres_exporter as non-superuser. Fixes #258.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,14 +203,10 @@ If you want to include only subset of databases, you can use option `--include-d
 
 ### Running as non-superuser
 
-To be able to collect metrics from `pg_stat_activity` and `pg_stat_replication`
-as  non-superuser you have to create functions and views as a superuser, and
-assign permissions separately to those.
-
-In PostgreSQL, views run with the permissions of the user that created them so
-they can act as security barriers. Functions need to be created to share this
-data with the non-superuser. Only creating the views will leave out the most
-important bits of data.
+To be able to collect metrics from `pg_stat*` views as non-superuser in PostgreSQL
+server versions >= 10 you can grant pg_monitor built-in role to the user. If
+you need to monitor older PostgreSQL servers, you will have to create functions
+and views as a superuser, and assign permissions separately to those.
 
 ```sql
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -203,14 +203,10 @@ If you want to include only subset of databases, you can use option `--include-d
...
-To be able to collect metrics from `pg_stat_activity` and `pg_stat_replication`
-as  non-superuser you have to create functions and views as a superuser, and
-assign permissions separately to those.
-
-In PostgreSQL, views run with the permissions of the user that created them so
-they can act as security barriers. Functions need to be created to share this
-data with the non-superuser. Only creating the views will leave out the most
-important bits of data.
+To be able to collect metrics from `pg_stat*` views as non-superuser in PostgreSQL
+server versions >= 10 you can grant pg_monitor built-in role to the user. If
```
mbanck, line 207: "you can grant *the* pg_monitor built-in role" I'd say, maybe also markup pg_monitor with backticks, like `pg_stat*` before.
Also, I guess technically the `pg_read_all_stats` role would suffice, so maybe mention both

juris-greitans: Fixed. PostgreSQL documentation says that `pg_monitor` is a member of `pg_read_all_stats`, that's why I went with `pg_monitor`.
Also, I added a link to pre-defined role documentation.

mbanck (approved):

runhardr:
Why is merging still blocked on this?

mbanck:
I guess my vote doesn't count for a lot

sysadmind (approved):

PR #592, opened 2021-12-01
Correctly set short_version for major version 10 and up

### LINKED ISSUE #349: short_version has wrong value for v10+
I understand that this (`11.6.0`, i.e the `.0` at the end) is mandated by semver, but then maybe it's time to ditch that and use something else. AFAIK semver assumed the `6` is the minor version while in fact it is the patch version for version 10 and up. Displaying `short_version` in dashboards (like in https://grafana.com/grafana/dashboards/6742 and more specifically https://grafana.com/api/dashboards/6742/images/4318/thumbnail) is pretty common, and displaying a  (albeit only slightly, but still) wrong version is jarring.

Please excuse me if this issue was brought up and rejected before, but I couldn't find it while searching through the issues.

mbanck:
Before, postgres_exporter reported a full three-digit semver for Postgres
versions in the pg_static metric, padding the version with a extra .0, e.g.
13.3 -> 13.3.0.

Fixes: https://github.com/prometheus-community/postgres_exporter/issues/349

### DIFF
--- a/cmd/postgres_exporter/postgres_exporter.go
+++ b/cmd/postgres_exporter/postgres_exporter.go
@@ -653,9 +653,15 @@ func (e *Exporter) checkMapVersions(ch chan<- prometheus.Metric, server *Server)
 	versionDesc := prometheus.NewDesc(fmt.Sprintf("%s_%s", namespace, staticLabelName),
 		"Version string as reported by postgres", []string{"version", "short_version"}, server.labels)
 
+	var semanticVersionShort = semanticVersion.String()
+	if semanticVersion.GE(semver.MustParse("10.0.0")) {
+		var versionSplit = strings.Split(semanticVersion.String(), ".")
+		semanticVersionShort = versionSplit[0] + "." + versionSplit[1]
+
+	}
 	if !e.disableDefaultMetrics && server.master {
 		ch <- prometheus.MustNewConstMetric(versionDesc,
-			prometheus.UntypedValue, 1, versionString, semanticVersion.String())
+			prometheus.UntypedValue, 1, versionString, semanticVersionShort)
 	}
 	return nil
 }


cmd/postgres_exporter/postgres_exporter.go
```diff
@@ -702,9 +702,15 @@ func (e *Exporter) checkMapVersions(ch chan<- prometheus.Metric, server *Server)
 	versionDesc := prometheus.NewDesc(fmt.Sprintf("%s_%s", namespace, staticLabelName),
 		"Version string as reported by postgres", []string{"version", "short_version"}, server.labels)
 
+	var short_version = semanticVersion.String()
```
SuperQ, line 705: This needs to follow Go naming conventions.

```suggestion
	var semanticVersionShort = semanticVersion.String()
```

SuperQ (changes requested):

cmd/postgres_exporter/postgres_exporter.go
mbanck: Changed this now in ce2f9089bd

mbanck:
On a general note, I think there should be an easily accessible metric for the versoin like
```
pg_version "15.3" 
```

SuperQ (approved):

cmd/postgres_exporter/postgres_exporter.go
```diff
@@ -653,9 +653,15 @@ func (e *Exporter) checkMapVersions(ch chan<- prometheus.Metric, server *Server)
 	versionDesc := prometheus.NewDesc(fmt.Sprintf("%s_%s", namespace, staticLabelName),
 		"Version string as reported by postgres", []string{"version", "short_version"}, server.labels)
 
+	var semanticVersionShort = semanticVersion.String()
+	if semanticVersion.GE(semver.MustParse("10.0.0")) {
+		var versionSplit = strings.Split(semanticVersion.String(), ".")
```
sysadmind, line 658: I wonder if checking for version 10 or greater could be fragile. What about always splitting on . but counting the parts instead? That seems safer to me and doesn't depend on the behavior of postgres.

sysadmind (approved):
I think it's okay the way it is but I think checking the parts count is better.

sysadmind:
@mbanck Can you please rebase this to clean up the conflicts?

ArthurSens:
Hi, since this PR seems to have been inactive for a very long time, I'm taking the liberty to close it.

Please do open it again if there's still work to be done here :)

mbanck:
Well, did this get fixed otherwise in the meantime?

ArthurSens:
It was not yet, if I understand correctly

mbanck:
Well ok then - I proposed this PR 5 years ago and I need to see whether I still have the branch around so I can rebase it in case you decide to reopen it.

Issue #358, opened 2020-01-07
Panic: descriptors reported by collector have inconsistent label names


amadeuszkryze:
Version:
`postgres_exporter_build_info{branch="",goversion="go1.11",revision="",version="0.0.1"} 1`

Error:
```bash
panic: descriptors reported by collector have inconsistent label names or help strings for the same fully-qualified name, offender is Desc{fqName: "pg_settings_effective_cache_size_bytes", help: "Sets the planner's assumption about the size of the data cache. [Units converted to bytes.]", constLabels: {server="some.server:5432"}, variableLabels: []}
```

Related issues:
#254 
#175 

Details:
Based on info in upper tickets its related to different postgres engines versions. But `disable-settings-metrics` is not an option here. 
I've plenty of postgres servers with different engine versions due to regressions and performance testing between them and I don't want to have plenty of exporters :D

rajalakshmi-v15:
Same issue here. I cannot disable settings since I am configuring alerts based on the preconfigured settings. Request you to please address the issue ASAP.

dgarcdu:
Hello.
Same issue here:

```
panic: descriptors reported by collector have inconsistent label names or help strings for the same fully-qualified name, offender is Desc{fqName: "pg_settings_google_insights_enabled", help: "Sets whether to enable the extension with shared memory allocation and stats recording.", constLabels: {server="172.22.36.243:5432"}, variableLabels: []}
```

Disabling pg_settings is not an option, as we have configured alerts based con max connections and others.

Also, deploying several exporters is not an option either, as it would bring some maintenance problems in the long term.

Could you please at least offer any workaround?

Thanks in advance, and best regards.

quenlang:
@dgarcdu 
I meet the same issue and the root cause is the different value of column short_desc in pg_settings table on different  postgres engines versions. prometheus not allow the same metric has different description when it was registered.
The short_desc of ```effective_cache_size_bytes``` which is different between 9.6.2 and 9.6.22
```
9.6.22: Sets the planner's assumption about the total size of the data caches. [Units converted to bytes.]
9.6.2: Sets the planner's assumption about the size of the disk cache. [Units converted to bytes.]
```

quenlang:
#598 fixed

sysadmind:
[...]

PR #598, opened 2021-12-26
fix issue #358: descriptors reported by collector have inconsistent l…

quenlang:
Cached the first server settings short description and the other server re-user the first server cache when there are multiple servers monitor.
The Prometheus metrics keep consistent in this way when they were registered.

Signed-off-by: zhoujk <quenlang@126.com>


sysadmind:
I'm not sure that I understand what this PR is trying to achieve. Can you provide a bit more context please?

ArthurSens:
Hey hey, I'm taking the liberty to close this PR since a solution was already merged in #771

PR #603, opened 2022-01-18
Document the custom query format

ringerc:
Add docs on the custom query format to the sample query yaml file.

I had to read the sources to properly understand the custom queries format, and I found the 'master' flag quite confusing. Hopefully these docs will help the next person.

I'm not sure if they should be rolled into the README.md, but I put them in the sample query yaml file for now.


SuperQ:
The custom queries.yaml is now deprecated.

PR #604, opened 2022-01-18
Clarify default values for CLI flags

alexrudd2:
Reference here: https://grafana.com/docs/agent/latest/configuration/integrations/postgres-exporter-config/
Also, matches the environment variables section.


sysadmind (approved):
LGTM

PR #613, opened 2022-02-10
Add pg_database collector

sysadmind:
Converts the pg_database metrics from queries.yaml to a built in collector. This is enabled by default because it is not likely to be a performance problem and likely very useful data.

### DIFF
--- a/cmd/postgres_exporter/server.go
+++ b/cmd/postgres_exporter/server.go
@@ -14,16 +14,13 @@
 package main
 
 import (
-	"context"
 	"database/sql"
 	"fmt"
-	"log"
 	"sync"
 	"time"
 
 	"github.com/blang/semver"
 	"github.com/go-kit/log/level"
-	"github.com/prometheus-community/postgres_exporter/collector"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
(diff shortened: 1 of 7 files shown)


cmd/postgres_exporter/server.go
```diff
@@ -128,6 +131,17 @@ func (s *Server) Scrape(ch chan<- prometheus.Metric, disableSettingsMetrics bool
 		err = fmt.Errorf("queryNamespaceMappings returned %d errors", len(errMap))
 	}
 
+	{
```
sysadmind, line 134: This is in a block to not reset the err var that gets set above. I don't want to return early because I don't think the errors are fatal, just informational.

collector/pg_database.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2021 The Prometheus Authors
```
SuperQ, line 1: 2022? :grin:

cmd/postgres_exporter/server.go
```diff
@@ -128,6 +131,17 @@ func (s *Server) Scrape(ch chan<- prometheus.Metric, disableSettingsMetrics bool
...
+		pgdb := collector.NewPGDatabaseCollector()
```
SuperQ, line 135: Rather than put every collector in the `main`, we should probably do what we do in the node_exporter.

So in `collector/pg_database.go`

```
func init() {
  registerCollector("pg_database", defaultDisabled, NewPGDatabaseCollector)
}
```

collector/pg_database.go
sysadmind: I'll get it right by next year

sysadmind:
@SuperQ Setting up the node_exporter style collector registration required a lot more code changes so this PR got a lot larger. I had to copy a lot of the node exporter functions. 

This will be a much better strategy moving forward however.

SuperQ (approved):
LGTM

PR #582, opened 2021-10-14
fix: command arguments

zilti:
It needs two dashes, otherwise, it won't work.

PR #614, opened 2022-02-10
Update rc file to reflect newer args

sysadmind:
This was not updated when the arguments were last updated. No need to list the options here IMO.  

I'm not positive if this file is actually used anywhere.


sysadmind:
Supersedes #582 (missing DCO)

PR #616, opened 2022-02-16
Add pg_database_size_bytes

fffergal:
We can see something similar in render dashboard, but want to collect for a longer time and see it in Grafana.

Query copied from upstream. Fun fact a week ago this was made a built in upstream, no query needed. Not released yet though.


fffergal:
Ah sorry, GitHub picked the wrong remote and I didn't notice. Ignore me.

PR #620, opened 2022-03-03
Sanitize setting values because of Aurora irregularity

### LINKED ISSUE #619: Aurora 13.5 returns shared_buffers setting with unit causing a panic
I have upgraded my AWS Aurora server to 13.5 today and I noticed that `postgres-exporter` started crashing.

From the logs, I see that the error is:
```
panic: Error converting setting "shared_buffers" value "88413056kB" to float: strconv.ParseFloat: parsing "88413056kB": invalid syntax
```

I looked at the code and I found a query selecting from `pg_settings`, so I tried it on the server, and indeed, `shared_buffers` is returning a value with the unit combined.

I tested the same query on a postgres in docker on my machine and 13.5 does not return this value with the unit, so this is something related to Aurora specifically.

I can work on a pull request. But I'm not sure what would be the best method. My thought was to have a test to see if the unit is included in the setting's value and strip it if it is the case.

This is the `shared_buffer` line from the query:
```
                        name                         | setting    | coalesce |                                                           short_desc                                                           | vartype
 shared_buffers                                      | 88413056kB | 8kB      | Sets the number of shared memory buffers used by the server.                                                                   | integer
```

SN9NV:
Check if a setting value ends with one of the known unit values and remove it.

This is because of a bug noticed in AWS RDS Aurora I ran into today when I upgraded to version 13.5

Fixes #619


angusburroughs (approved):

tibimoise:
We also need this fix. Thanks @SN9NV

sysadmind (approved):

PR #624, opened 2022-03-23
Add --strip-settings-suffix flag for #619

sysadmind:
Adds --strip-settings-suffix flag to make the fixes for AWS Aurora in #619 opt in for users. It should be safe for all users, but this code should also be unnecessary if AWS fixes their bug.


sysadmind:
We determined that this shouldn't hurt to leave in for all users.

PR #625, opened 2022-03-23
Update queries.yaml

ctouil:


sysadmind:
It's not clear what the intent of this PR is. The queries.yaml file is being deprecated in the near future.

benjimin:
The intent of this PR is evidently to update the syntax from postgres 12 (and older) to postgres 13+ (noting postgres 15 is current now).

https://www.postgresql.org/docs/current/pgstatstatements.html

mcosta74:
> It's not clear what the intent of this PR is. The queries.yaml file is being deprecated in the near future.

what does it mean? are you planning to remove the ability to run extra queries?

mcosta74:
> > what does it mean? are you planning to remove the ability to run extra queries?
> 
> Correct. We plan to remove this feature entirely as we migrate each postgres feature collector to Go code. By moving the queries into Go, we can better support multiple versions of PostgreSQL automatically. This is how it's done in the MySQL exporter and it is a far better user experience.
> 
> Most of the use of queries.yaml (business metrics, not PostgreSQL server metrics) is actually better suited for the more generic [sql_exporter](https://github.com/burningalchemist/sql_exporter).

So you're suggesting to **NOT** use custom queries but one instance of the `sql_exporter` to run business queries?

sysadmind:
@mcosta74 That is correct.

mcosta74:
Thanks for the info. I'm trying sql_exporter and it's exactly what I was
looking for.

==============

Massimo Costa

Il mar 9 mag 2023, 10:42 Joe Adams ***@***.***> ha scritto:

SuperQ:
queries.yaml is now deprecated. Updates should now go into the `collector` package.

PR #618, opened 2022-02-25
Add multi-target support

sysadmind:
- Remove multi server support from new collector package
- Add http handler for multi-target support

Currently WIP. 

TODO: 
[ ] Add auth module
[ ] Determine options for collecting the metrics
[ ] Remove existing multi-dsn support

Fixes: https://github.com/prometheus-community/postgres_exporter/issues/487

Signed-off-by: Joe Adams <github@joeadams.io>

PR #632, opened 2022-04-13
fix: default value of datasource when no environment variables

yangliyl:
https://github.com/prometheus-community/postgres_exporter/blob/92bdb8755de886b2366de245e559538030ba3668/cmd/postgres_exporter/main.go#L88-L97

default value of datasource when no environment variables
`[postgresql://:@]`

check is always false


ArthurSens:
Hey, sorry for the delayed review here. Although this PR probably made sense years ago, this exporter now supports the multi-target pattern (#618), so it is totally normal to start the exporter without a database configured.

That said, I'm closing this PR. Thanks!

PR #639, opened 2022-04-27
Improve PostgreSQL replication lag detection

wit:
If there is no updates in the database, the lag will not grow.

Signed-off-by: WiT <witphg@gmail.com>


SuperQ:
This has been migrated to the `collector` package. See /collector/pg_replication.go.

PR #677, opened 2022-08-18
Sanitize url for printing when given password passed as parameter

### LINKED ISSUE #643: Credentials passed as as url parameters are visible in log output
**What did you do?**

I set DATA_SOURCE_NAME with connection string providing user name and password as url parameters.
This is the format of my connection DATA_SOURCE_NAME string:  
postgresql://<postgres dns>:<port>/<database>?user=<username>&password=<password>

I populated the user and password parameters with invalid credentials. Bad password, etc.
Then I ran postgres_exporter application and loaded my 0.0.0.0/metrics endpoint

In my logging output, I expected to see "PASSWORD REMOVED" in message text. 

Instead the full password is logged.

I can work around this issue by changing my connection string to be formatted as
postgresql://<username>:<password>@<postgres dns>:<port>/<database>

but it seems like loggableDSN() function should handle this condition as well.

* System information:

	Darwin 21.4.0 arm64

* postgres_exporter version:

	version 0.10.1

* postgres_exporter flags:

        --web.listen-address=":8080"

AlexDCraig:
Resolves https://github.com/prometheus-community/postgres_exporter/issues/643

### DIFF
--- a/cmd/postgres_exporter/util.go
+++ b/cmd/postgres_exporter/util.go
@@ -217,5 +217,14 @@ func loggableDSN(dsn string) string {
 		pDSN.User = url.UserPassword(pDSN.User.Username(), "PASSWORD_REMOVED")
 	}
 
+	// Blank password data from parameters if not nil
+	q := pDSN.Query()
+	if q != nil {
+		if q.Get("password") != "" {
+			q.Set("password", "PASSWORD_REMOVED")
+			pDSN.RawQuery = q.Encode()
+		}
+	}
+
 	return pDSN.String()
 }


SuperQ:
We shouldn't be accepting `password` as a scrape URL parameter. Prometheus does not protect passwords in the URL params.

ArthurSens:
Hey, sorry for the delayed review. The PR is quite old, so it will need a rebase, but the bug still seems to exist. I added reproduction steps in #643.

Could you rebase the PR?

sysadmind:
I think #1389 supersedes this change. This older function isn't worth investing in because we're moving away from it.

PR #678, opened 2022-08-25
Add dsn type for handling datasources

sysadmind:
dsn is designed to replace the other uses of dsn as a string in the long term. dsn is designed to be safe to log, properly redacting passwords. The goal is eventually always parse datasource information into a dsn type object which can safely be passed around and logged without worrying about wrapping calls in a redaction function (today this function is loggableDSN().

This should solve the root issue in #648, #677, and #643, although the full fix will require more changes to update all code references over to use the dsn type.

### DIFF
--- a/cmd/postgres_exporter/datasource.go
+++ b/cmd/postgres_exporter/datasource.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"unicode"
 
 	"github.com/go-kit/log/level"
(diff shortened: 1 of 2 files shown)


sysadmind:
@SuperQ I wanted to get this base implementation in for review before starting to migrate code references.

SuperQ (approved):
Nice. This would be a nice upstream contribution to lib/pq.

Similar to the MySQL driver Config.

https://pkg.go.dev/github.com/go-sql-driver/mysql?#Config

PR #683, opened 2022-09-01
Extending /probe path metrics

preved911:
Hello. I saw that `/probe` path doesn't execute queries from file, doesn't collect different metrics, which are exist over default `/metrics` path and so on. I think, that it is unclear and unexpected behaviour. Could you please review this and tell me what I can do yet to merge it?


preved911:
Will it be reviewed?

sysadmind (approved):
LGTM. Thanks

Issue #588, opened 2021-11-12
exclude-databases does not properly work


psvampa:
**What did you do?**
I ran an exporter process including the auto-discover-databases and exclude-databases flags. Also pulling from a custom query files:
```
exec postgres_exporter --web.listen-address :9189 --disable-default-metrics --auto-discover-databases --exclude-databases postgres,template0,template1 --extend.query-path /tmp/queries.yaml
``` 
/tmp/queries.yaml file includes two different queries:
```
    pg_database:
      query: "SELECT pg_database.datname, pg_database_size(pg_database.datname) as size_bytes FROM pg_database"
      master: true
      metrics:
        - datname:
            usage: "LABEL"
            description: "Name of the database"
        - size_bytes:
            usage: "GAUGE"
            description: "Disk space used by the database"

    pg_database_2:
      query: "SELECT pg_database.datname, pg_database_size(pg_database.datname) as size_bytes FROM pg_database WHERE datname = current_database()"
      master: true
      metrics:
        - datname:
            usage: "LABEL"
            description: "Name of the database"
        - size_bytes:
            usage: "GAUGE"
            description: "Disk space used by the database"
```
DBs on my postgres instance:
```
postgres=# \l
                                 List of databases
   Name    |  Owner   | Encoding |  Collate   |   Ctype    |   Access privileges
-----------+----------+----------+------------+------------+-----------------------
 nsoengas  | postgres | UTF8     | en_US.utf8 | en_US.utf8 | =Tc/postgres         +
           |          |          |            |            | postgres=CTc/postgres+
           |          |          |            |            | nsoengas=C/postgres
 postgres  | postgres | UTF8     | en_US.utf8 | en_US.utf8 |
 template0 | postgres | UTF8     | en_US.utf8 | en_US.utf8 | =c/postgres          +
           |          |          |            |            | postgres=CTc/postgres
 template1 | postgres | UTF8     | en_US.utf8 | en_US.utf8 | =c/postgres          +
           |          |          |            |            | postgres=CTc/postgres
(4 rows)

postgres=#
```

**What did you expect to see?**
Since I am using exclude-databases flag and then excluding postgres,template0,template1 DBs, then my expectation is postgres exporter connecting to nsoengas DB and performing the queries on my custom file (/tmp/queries.yaml).
According to those queries, pg_database should retrieve datname and size for every single database on given instance. pg_database_2 should **ONLY** show datname and size for the database I am connected to. I've included the proper query filter to do that. 
[...]

PR #697, opened 2022-10-04
Fix exclude-databases for collector package

sysadmind:
The pg_database collector was not respecting the --exclude-databases flag and causing problems where databases were not accessible. This now respects the list of databases to exclude.

- Adjusts the Collector create func to take a config struct instead of a logger. This allows more changes like this in the future. I figured we would need to do this at some point but I wasn't sure if we could hold off.
- Split the database size collection to a separate query when database is not excluded.
- Comment some probe code that was not useful/accurate

### DIFF
--- a/cmd/postgres_exporter/probe.go
+++ b/cmd/postgres_exporter/probe.go
@@ -61,21 +61,9 @@ func handleProbe(logger log.Logger, excludeDatabases []string) http.HandlerFunc 
 
 		// TODO(@sysadmind): Timeout
 
-		// probeSuccessGauge := prometheus.NewGauge(prometheus.GaugeOpts{
-		// 	Name: "probe_success",
-		// 	Help: "Displays whether or not the probe was a success",
-		// })
-		// probeDurationGauge := prometheus.NewGauge(prometheus.GaugeOpts{
-		// 	Name: "probe_duration_seconds",
-		// 	Help: "Returns how long the probe took to complete in seconds",
-		// })
-
 		tl := log.With(logger, "target", target)
 
-		// start := time.Now()
 		registry := prometheus.NewRegistry()
-		// registry.MustRegister(probeSuccessGauge)
-		// registry.MustRegister(probeDurationGauge)
 
 		opts := []ExporterOpt{
 			DisableDefaultMetrics(*disableDefaultMetrics),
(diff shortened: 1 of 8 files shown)


sysadmind:
also, this fixes #588

cmd/postgres_exporter/probe.go
```diff
@@ -62,27 +61,27 @@ func handleProbe(logger log.Logger) http.HandlerFunc {
...
-			Help: "Returns how long the probe took to complete in seconds",
-		})
+		// probeSuccessGauge := prometheus.NewGauge(prometheus.GaugeOpts{
+		// 	Name: "probe_success",
+		// 	Help: "Displays whether or not the probe was a success",
+		// })
+		// probeDurationGauge := prometheus.NewGauge(prometheus.GaugeOpts{
+		// 	Name: "probe_duration_seconds",
+		// 	Help: "Returns how long the probe took to complete in seconds",
+		// })
```
SuperQ, line 71: Let's just delete this. We don't need to expose this stuff.

We already have a `postgres_up` metric.

sysadmind:
@SuperQ Good idea. I removed the commented code.

macmiranda:
@sysadmind you planning on fixing the conflicts?

cmd/postgres_exporter/probe.go
```diff
@@ -96,10 +83,10 @@ func handleProbe(logger log.Logger) http.HandlerFunc {
 		registry.MustRegister(exporter)
 
 		// Run the probe
-		pc, err := collector.NewProbeCollector(tl, registry, dsn)
+		pc, err := collector.NewProbeCollector(tl, excludeDatabases, registry, dsn)
 		if err != nil {
-			probeSuccessGauge.Set(0)
-			probeDurationGauge.Set(time.Since(start).Seconds())
+			// probeSuccessGauge.Set(0)
+			// probeDurationGauge.Set(time.Since(start).Seconds())
```
SuperQ, line 89:
```suggestion
```

SuperQ (changes requested):
One more remove dead code.

cmd/postgres_exporter/probe.go
sysadmind: Good catch. I missed it in the rebase. Thanks. All fixed up now (I hope)

SuperQ (approved):

PR #702, opened 2022-10-12
Update mixin alerts summary annotation

v-zhuravlev:
Update mixin alerts summary annotation to use simple short strings.

Expanded description (what happened, where?) can be found in 'description' annotation instead, while `summary` is better used as Human readable alert name.  

See also https://monitoring.mixins.dev/#guidelines-for-alert-names-labels-and-annotations


sysadmind:
@SuperQ What are your thoughts? I'm not that familiar with mixins. I'm also not sure what these changes would mean for users or if the changes are important.

ArthurSens:
Hey, sorry for the delayed review here. It looks like #1179 already solved the same problem, so I'm closing this one. Thanks!

PR #703, opened 2022-10-13
Fix templated variables in postgres-mixin

v-zhuravlev:
- Make sure that the dash would work with any datasource name.
- Change 'instance' var query to use pg_up metric instead of up
- Change order of vars: datasource, job, instance, db.

Signed-off-by: Vitaly Zhuravlev <zhuravlev.vitaly@gmail.com>


ArthurSens:
Hi, sorry for the super-delayed review here. Seems like #1179 already solved most of the problems this PR is also solving.

I'm taking the liberty of closing this one

Issue #727, opened 2022-12-01
postgres_exporter left more and more conntions in database and not released or reuse them 


sheenzxx:
when use postgres_exporter to monitor pg database,

postgres_exporter left more and more connections in database and not released or reuse them

![1669889809(1)](https://user-images.githubusercontent.com/3350015/205027177-5bf8a7ac-278c-4b67-8498-f7c2cac78fe2.png)

then on other windows to request metrics

curl "http://localhost:9187/probe?auth_module=foo1&target=192.168.74.94:5433/pcauto"

query 
select datid,pid,usename,query_start from pg_stat_activity where usename='exporter_monitor' order by query_start

can see that when request 1 time .1 more connection add and not released or reuse
![1669889957(1)](https://user-images.githubusercontent.com/3350015/205027616-cce30d87-5833-496e-84ed-6f48929e7164.png)

sheenzxx:
if there any Perfect solution？

hajdukda:
ok this one is super important to fix.

kjbass:
Yeah, another +1 for this being an important one to fix.  Running the exporter caused number of connections to go up endlessly.  @sheenzxx's solution seems like a valid one, but another option would be actually closing the connection after a probe finishes.

kjbass:
> Yeah, another +1 for this being an important one to fix. Running the exporter caused number of connections to go up endlessly. @sheenzxx's solution seems like a valid one, but another option would be actually closing the connection after a probe finishes. Currently, cannot really use this in prod.

I might try my hand at fixing this one, think this shouldn't be too terrible to auto-close after probing.

kjbass:
Tried my hand at writing a fix here: https://github.com/prometheus-community/postgres_exporter/pull/734

kazauwa:
This bug can easily bring the production database down. This is what we saw yesterday after enabling `postgres_exporter`. We definitely need [https://github.com/prometheus-community/postgres_exporter/pull/734](734) merged.
<img width="783" alt="Screenshot 2023-02-01 at 12 22 38" src="https://user-images.githubusercontent.com/12330159/215946803-85cf7320-d640-4995-bcb9-29b8a09f4ddb.png">

MichaelDBA:
Doesn't the error message indicate its a configuration problem?  --> error = empty dsn

So perhaps the connection is made nonetheless but unawares to the postgres_exporter so that's the reason its never closed.

kjbass:
[...]

PR #734, opened 2022-12-14
Close DB connection after probe completes.

kjbass:
Currently each probe creates a new DB connection which is never closed.  This fix closes the connection upon the completion of a probe.  An alternative solution would be to inject a DB object into the handleProbe handler, but that would be a larger scope change.

Tested locally and scrapes correctly without increasing the number of connections in Postgres.  

edit: since opening the PR I have been running this in a dev environment and scrapes are working as expected and connection level is staying stable.

Will fix https://github.com/prometheus-community/postgres_exporter/issues/727

Signed-off-by: Kurtis Bass <kurtis.bass@hinge.co>

### DIFF
--- a/cmd/postgres_exporter/probe.go
+++ b/cmd/postgres_exporter/probe.go
@@ -104,6 +104,9 @@ func handleProbe(logger log.Logger) http.HandlerFunc {
 			return
 		}
 
+		// Cleanup underlying connections to prevent connection leaks
+		defer pc.Close()
+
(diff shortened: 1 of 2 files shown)


nthienan:
Can anybody process this pr?

nthienan:
@SuperQ Can you help to review this?

sysadmind (approved):
LGTM Thanks!

PR #744, opened 2023-01-20
adding replication slot metric

sheldor14:
We would like to have metrics based on the distance between the current wal lsn and the confirmed flushed lsn on replication slots. The PR adds that metric to the queries.yaml file


sheldor14:
cc @SuperQ

sysadmind:
You can freely edit your own queries.yaml file as you wish. The one in this repo is just an example of possible options.

sheldor14:
> You can freely edit your own queries.yaml file as you wish. The one in this repo is just an example of possible options.

omg you're totally right. i completely misread the usage haha. thanks for the heads up!

PR #746, opened 2023-01-24
Update exporter-toolkit

SuperQ:
Update to the latest exporter-toolkit
* Enables multi-listener and systemd socket activation.
* Bump Go to 1.19.
* Remove `PG_EXPORTER_WEB_LISTEN_ADDRESS` env var because this is now a repeatable flag.

Signed-off-by: SuperQ <superq@gmail.com>


cmd/postgres_exporter/main.go
```diff
@@ -146,9 +145,8 @@ func main() {
 
 	http.HandleFunc("/probe", handleProbe(logger))
 
-	level.Info(logger).Log("msg", "Listening on address", "address", *listenAddress)
```
sysadmind, line 149: Does the exporter toolkit log the listen address? That seems useful for users. Cuts down on issues filed.

sysadmind (approved):
LGTM.

cmd/postgres_exporter/main.go
SuperQ: Yup, the new version of the toolkit handles the logging, since the listen address can be called multiple times.

yukha-dw:
Hello @SuperQ, is there any way to use custom port using environment variable after this change?

Thank you in advance!

PR #747, opened 2023-01-24
adding codified functionality for logical replication metrics

sheldor14:
given the push to codify some of the queries.yaml statements, we want to add logical replication metrics. This metric will be labeled on the slot name and the metric will be the distance between the `current_wal_lsn()` and the confirmed_flush_lsn

### DIFF
--- a/collector/pg_replication_slots.go
+++ b/collector/pg_replication_slots.go
@@ -1,0 +1,72 @@
+// Copyright 2022 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package collector
+
+import (
+	"context"
+	"database/sql"
+
+	"github.com/go-kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func init() {
+	registerCollector("replication_slot", defaultEnabled, NewPGReplicationSlotCollector)
+}
+
+type PGReplicationSlotCollector struct {
+	log log.Logger
+}
+
+func NewPGReplicationSlotCollector(logger log.Logger) (Collector, error) {
+	return &PGReplicationSlotCollector{log: logger}, nil
+}
+
+var pgReplicationSlot = map[string]*prometheus.Desc{
+	"lsn_distance": prometheus.NewDesc(
+		"pg_replication_slot_lsn_distance",
+		"Disk space used by the database",
+		[]string{"slot_name"}, nil,
+	),
+}
+
+func (PGReplicationSlotCollector) Update(ctx context.Context, db *sql.DB, ch chan<- prometheus.Metric) error {
+	rows, err := db.QueryContext(ctx,
+		`SELECT
+			slot_name,
+			(pg_current_wal_lsn() - confirmed_flush_lsn) AS lsn_distance
+			FROM
+		pg_replication_slots;`)
+	if err != nil {
+		return err
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		var slot_name string
+		var size int64
+		if err := rows.Scan(&slot_name, &size); err != nil {
+			return err
+		}
+
+		ch <- prometheus.MustNewConstMetric(
+			pgReplicationSlot["size_bytes"],
+			prometheus.GaugeValue, float64(size), slot_name,
+		)
+	}
+	if err := rows.Err(); err != nil {
+		return err
+	}
+	return nil
+}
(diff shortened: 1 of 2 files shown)


sheldor14:
cc @SuperQ

collector/pg_replication_slots.go
```diff
@@ -0,0 +1,72 @@
...
+		"Disk space used by the database",
+		[]string{"slot_name"}, nil,
+	),
+}
+
+func (PGReplicationSlotCollector) Update(ctx context.Context, db *sql.DB, ch chan<- prometheus.Metric) error {
+	rows, err := db.QueryContext(ctx,
+		`SELECT
+			slot_name,
+			(pg_current_wal_lsn() - confirmed_flush_lsn) AS lsn_distance
```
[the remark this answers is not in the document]
sheldor14: fair point. happy to add the raw values. would be easier to debug any issues later and then we can just do the math afterwards

collector/pg_replication_slots.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"context"
+	"database/sql"
+
+	"github.com/go-kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func init() {
+	registerCollector("replication_slot", defaultEnabled, NewPGReplicationSlotCollector)
```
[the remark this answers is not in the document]
sysadmind: I don't know that I have a strong opinion here. The tables inside the database have the pg_ prefix, so I could argue keeping the prefix. I'm not opposed to either direction.

collector/pg_replication_slots.go
```diff
@@ -0,0 +1,72 @@
...
+
+type PGReplicationSlotCollector struct {
+	log log.Logger
+}
+
+func NewPGReplicationSlotCollector(logger log.Logger) (Collector, error) {
+	return &PGReplicationSlotCollector{log: logger}, nil
+}
+
+var pgReplicationSlot = map[string]*prometheus.Desc{
```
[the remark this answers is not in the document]
sysadmind: The thing about putting them in the struct of each collector is that it makes the New func a bit bloated in my opinion. Putting them in the struct would keep the package namespace more clean though. If they're in the struct, the probe endpoint will recreate them on every request the way that it is coded today, so that might increase the memory footprint of the collector. I think if we change it to struct scope that we should do that at a later time after we refactor the probe endpoint to not re-init the collector package on every request.

collector/pg_replication_slots.go
```diff
@@ -0,0 +1,83 @@
...
+		pg_replication_slots;`)
+	if err != nil {
+		return err
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		var slot_name string
+		var wal_lsn int64
+		var flush_lsn int64
```
sysadmind, line 65: The docs say that this could be null so this should be a sql.NullInt64 to account for that. I'm not sure what the value should be in that case. I suppose 0 would be fine. 

https://www.postgresql.org/docs/current/view-pg-replication-slots.html

collector/pg_replication_slots.go
```diff
@@ -0,0 +1,83 @@
...
+		if err := rows.Scan(&slot_name, &wal_lsn, &flush_lsn); err != nil {
+			return err
+		}
+
+		ch <- prometheus.MustNewConstMetric(
+			pgReplicationSlot["current_wal_lsn"],
+			prometheus.GaugeValue, float64(wal_lsn), slot_name,
```
sysadmind, line 72: Should these be a gauge? Won't this value only increase?

sysadmind (changes requested):
Overall it looks good to me. Just an edge case to cover and I'm unsure of Gauge vs Counter.

collector/pg_replication_slots.go
sheldor14: i agree with you that it should be 0, i'll also add a bool counter for is_active to denote whether or not the 0 is actually useful here.

collector/pg_replication_slots.go
sheldor14: counter sounds good to me. something i think i also need to do is that these values are in pg_lsn. im thinking in the sql statement do the `pg_lsn - - '0/0'` to get an actual int value. now for the null i can do `coalesce(confirmed_flush_lsn, '0/0') - '0/0'`

collector/pg_replication_slots.go
SuperQ: Usually for null, we don't expose the metric.

"is_active" should probably be a 0/1 boolean gauge like `pg_replication_lsn_active` or something like that.

collector/pg_replication_slots.go
SuperQ: Depends on if this is a truly a counter in the underlying system.

collector/pg_replication_slots.go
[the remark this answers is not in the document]
SuperQ: Since this is local to the package, I think we should name them after the collector to keep with the convention. `pg_` seems redundant.

collector/pg_replication_slots.go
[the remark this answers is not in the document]
SuperQ: >  If they're in the struct, the probe endpoint will recreate them on every request the way that it is coded today, so that might increase the memory footprint of the collector.

Ahh, right, that's a good reason to keep them as package vars outside of the struct. Reduced GC churn.

SuperQ (approved):
Thanks!

SuperQ:
Ping @sysadmind

sysadmind (approved):
LGTM

Issue #754, opened 2023-02-07
FreeBSD: send stdout to syslog


gavrik:
It is a good practise send stdout and stderr to syslog and add syslog_tag to this logs.

sysadmind:
Fixed in #755

PR #755, opened 2023-02-07
send stdout/stderr to syslog

gavrik:
Fix for issue #754 
Adding to parameters for **daemon**
* -f: redirect output
* -T: add syslog_tag
Signed-off-by: Mike <gavrikster@gmail.com>


sysadmind (approved):
I'm not familiar with the BSD .rc files so I am just assuming that this works.

PR #761, opened 2023-02-28
Remove QueryID to fix #549

### LINKED ISSUE #549: `pg_stat_statement` metrics have unreasonable cardinality

benjimin:
If users follow the RDS documentation included in this repo, they will use this SQL yaml and encounter a cardinality explosion, because 19 separate metrics are stored for each new query that gets performed. Removed `queryid` to fix #549.

### DIFF
--- a/queries.yaml
+++ b/queries.yaml
@@ -148,7 +148,7 @@ pg_statio_user_tables:
 
 # WARNING: This set of metrics can be very expensive on a busy server as every unique query executed will create an additional time series
 pg_stat_statements:
-  query: "SELECT t2.rolname, t3.datname, queryid, calls, total_time / 1000 as total_time_seconds, min_time / 1000 as min_time_seconds, max_time / 1000 as max_time_seconds, mean_time / 1000 as mean_time_seconds, stddev_time / 1000 as stddev_time_seconds, rows, shared_blks_hit, shared_blks_read, shared_blks_dirtied, shared_blks_written, local_blks_hit, local_blks_read, local_blks_dirtied, local_blks_written, temp_blks_read, temp_blks_written, blk_read_time / 1000 as blk_read_time_seconds, blk_write_time / 1000 as blk_write_time_seconds FROM pg_stat_statements t1 JOIN pg_roles t2 ON (t1.userid=t2.oid) JOIN pg_database t3 ON (t1.dbid=t3.oid) WHERE t2.rolname != 'rdsadmin'"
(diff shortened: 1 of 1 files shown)


sysadmind:
@SuperQ What are your thoughts on this?

SuperQ:
The query id is the primary reason that this collector exists, so, not really useful without it.

benjimin:
@SuperQ could you share a bit of explanation for _why_ you want the query id in **metrics** (as opposed to **logs**)? 

(What would the benefit be, of exporting a separate time series for each query id - given that the query id generally will never reoccur and thus the values in the time series will never fluctuate?)

This kind of flies directly in the face of Prometheus design and best practice. See, for example, the admonition from the [prometheus documentation](https://prometheus.io/docs/practices/naming/#labels):

| CAUTION: 
|----------
| _Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values._

At minimum, do you think it would be prudent to update `README-RDS.md` to warn the user about the consequences of using the `query.yaml` in the manner currently advised (particularly since there are also significant financial risks noted in the #549 issue discussion)?

SuperQ:
Because in many applications, the number of query ID patterns is fairly limited, and the number of database instance is also much more constrained compared to the number of application servers. This makes the total cardinality acceptable for a lot of use cases.

Knowing the about the performance characteristics of the top X queries can be extremely useful, which is why it's in the example file.

We're planning to remove support for the `queries.yaml` by moving the various queries into Go code. The whole feature is an anti-pattern that was added to this exporter before it was adopted by the community. As you noticed, people blindly copy the file into production.

 I don't think this has anything to do with RDS or not. Any application can produce a lot of series.

What would be OK with me would be to simply comment out the whole query in the yaml file. This way it's not on by default for users that are prone to copy-pasta and don't read through the file to see what it does. I also made some suggested changes to the query but never bothered to update it. That's one of the reasons why we want to eliminate this example file. Good improvements get made locally and never make it upstream.

PR #764, opened 2023-03-05
Update build

SuperQ:
* Update Go to 1.20.
* Update golanci-lint.
* Bump modules.
* Update CI orb.


.github/workflows/golangci-lint.yml
```diff
@@ -20,11 +20,11 @@ jobs:
       - name: install Go
         uses: actions/setup-go@v2
```
sysadmind, line 21: Looks like we can upgrade this to `v3` as well.

.github/workflows/golangci-lint.yml
```diff
@@ -20,11 +20,11 @@ jobs:
...
         with:
-          go-version: 1.18.x
+          go-version: 1.20.x
       - name: Install snmp_exporter/generator dependencies
         run: sudo apt-get update && sudo apt-get -y install libsnmp-dev
         if: github.repository == 'prometheus/snmp_exporter'
       - name: Lint
-        uses: golangci/golangci-lint-action@v3.2.0
```
sysadmind, line 28: What is the reason that we are explicitly calling out the full version here? We could use `v3` and each run should reference the latest. I assume we had some issue in the past?

sysadmind (commented):
Looks like we may need to fix the lint issue here too.

.github/workflows/golangci-lint.yml
SuperQ: Good question, I've always seen these use full semver. I just recently tried to use a major version like that in another action, but it failed to work without full semver.

.github/workflows/golangci-lint.yml
SuperQ: We used to sync these files from the prometheus repo. But we're slowly converting away from CircleCI to just GH Actions. I'm trying to do the minimal work here to keep these up to date until the full conversion (and return to syncing) is done.

SuperQ:
Ok, fixed the linting issues.

sysadmind (approved):
LGTM

PR #766, opened 2023-03-08
Automatically set GOMAXPROCS based on CPU quota to avoid throttling

yann-soubeyrand:
Hello,

In Go, maximum parallelism can be controlled via the GOMAXPROCS environment variable. If not set, this setting defaults to the number of CPUs available on the machine even if the program is running in a container with CPU quota: https://github.com/golang/go/issues/33803.

Due to the functioning of the CFS scheduler in Linux kernel, this can lead to CPU throttling even if CPU usage is well below the quota (here the quota is a full CPU, 1000 millicores):

**CPU usage in millicores**

CPU usage in millicores

**CPU throttling in percentage**

CPU throttling in percentage

To solve this problem, Uber maintains a package to automatically set GOMAXPROCS based on CPU quota: https://github.com/uber-go/automaxprocs. When using it, the situation greatly improves (still with the same CPU quota as above):

**CPU usage in millicores**

CPU usage in millicores

**CPU throttling in percentage**

CPU throttling in percentage

This PR proposes to use the automaxprocs package from Uber.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ auth_modules:
 To build the Docker image:
 
     make promu
-    promu crossbuild -p linux/amd64 -p linux/armv7 -p linux/amd64 -p linux/ppc64le
+    promu crossbuild -p linux/amd64 -p linux/armv7 -p linux/arm64 -p linux/ppc64le
     make docker
 
 This will build the docker image as `prometheuscommunity/postgres_exporter:${branch}`.
--- a/cmd/postgres_exporter/main.go
+++ b/cmd/postgres_exporter/main.go
@@ -125,7 +125,7 @@ func main() {
 	}
 
 	excludedDatabases := strings.Split(*excludeDatabases, ",")
-	logger.Log("msg", "Excluded databases", "databases", fmt.Sprintf("%v", excludedDatabases))
+	level.Info(logger).Log("msg", "Excluded databases", "databases", fmt.Sprintf("%v", excludedDatabases))
 
 	opts := []ExporterOpt{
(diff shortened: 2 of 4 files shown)


SuperQ:
We added some experimental support for this to prometheus/prometheus.

I'm wondering if we should add this boilerplate to the [exporter-toolkit](https://github.com/prometheus/exporter-toolkit) to reduce the amount of copy-n-paste we have in the exporters.

We could have a package like 

```
import (
  _ "https://github.com/prometheus/exporter-toolkit/automaxprocs"
)
```

The only real down side to the way this package works is that it doesn't understand Kubernetes-style requests which allow for better binpacking than hard CFS limits.

yann-soubeyrand:
> I'm wondering if we should add this boilerplate to the [exporter-toolkit](https://github.com/prometheus/exporter-toolkit) to reduce the amount of copy-n-paste we have in the exporters. CC @roidelapluie

That’s seems to be a good idea to me.

yann-soubeyrand:
Hello @SuperQ @roidelapluie, I’m having a look at the exporter toolkit to add `automaxprocs` to it, but I’m not sure how you’d like to integrate it. Should I put it in an `init()` in the web package or should I better create a `maxprocs` package with this `init()`?

evevseev:
Hi! I am facing the same problem of throttling. Do you have an ETA for merging this PR or re-implementing using exporter toolkit?
Thanks

SuperQ:
@evevseev

No ETA. IMO it's just easier to set `GOMAXPROCS=1` in your environment. Exporters typically need almost no CPU. So it's not a complicated thing that needs a lot of tuning.

Issue #770, opened 2023-03-13
pg_settings metrics have different short descriptions in different pg versions


khkhalifa:
**What did you do?**

Configured postgres exporter with multiple targets having different versions (13.9.0 & 14.6.0)
and running the /metrics endpoint

**What did you expect to see?**

To get the metrics

**What did you see instead? Under which circumstances?**

Multiple errors

* postgres_exporter version:

	insert output of `postgres_exporter --version` here

* postgres_exporter flags:

--auto-discover-databases
--web.listen-address=:9187

* PostgreSQL version:

	13.9.0 & 14.6.0

* Logs:
```
insert logs relevant to the issue here
```

The response from the metrics endpoint

An error has occurred while serving metrics:

11 error(s) occurred:
* collected metric pg_settings_check_function_bodies label:<name:"server" value:"<server>:5432" > gauge:<value: 0>  has help "Check function bodies during CREATE FUNCTION." but should have "Check routine bodies during CREATE FUNCTION and CREATE PROCEDURE."
* collected metric pg_settings_data_directory_mode label:<name:"server" value:"<server>:5432" > gauge:<value: 0>  has help "Mode of the data directory." but should have "Shows the mode of the data directory."
* collected metric pg_settings_enable_partition_pruning label:<name:"server" value:"<server>:5432" > gauge:<value: 0>  has help "Enables plan-time and run-time partition pruning." but should have "Enables plan-time and execution-time partition pruning."
* collected metric pg_settings_idle_in_transaction_session_timeout_seconds label:<name:"server" value:"<server>:5432" > gauge:<value: 0>  has help "Sets the maximum allowed duration of any idling transaction. [Units converted to seconds.]" but should have "Sets the maximum allowed idle time between queries, when in a transaction. [Units converted to seconds.]"
* collected metric pg_settings_integer_datetimes label:<name:"server" value:"<server>:5432" > gauge:<value: 0>  has help "Datetimes are integer based." but should have "Shows whether datetimes are integer based."
* collected metric pg_settings_jit_debugging_support label:<name:"server" value:"<server>:5432" > gauge:<value: 0>  has help "Register JIT compiled function with debugger." but should have "Register JIT-compiled functions with debugger."
* collected metric pg_settings_jit_optimize_above_cost label:<name:"server" value:"<server>:5432" > gauge:<value: 0>  has help "Optimize JITed functions if query is more expensive." but should have "Optimize JIT-compiled functions if query is more expensive."
[...]

PR #771, opened 2023-03-13
Fix pg_setting different help values

khkhalifa:
pg_settings metrics have different short descriptions in different pg versions #770

Similar to #598

### DIFF
--- a/cmd/postgres_exporter/pg_setting.go
+++ b/cmd/postgres_exporter/pg_setting.go
@@ -70,7 +70,7 @@ func (s *pgSetting) metric(labels prometheus.Labels) prometheus.Metric {
 		err       error
 		name      = strings.Replace(s.name, ".", "_", -1)
 		unit      = s.unit // nolint: ineffassign
-		shortDesc = s.shortDesc
+		shortDesc = fmt.Sprintf("Server Parameter: %s", s.name)
 		subsystem = "settings"
 		val       float64
 	)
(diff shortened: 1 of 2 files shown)


khkhalifa:
cc @SuperQ @sysadmind

cmd/postgres_exporter/pg_setting.go
```diff
@@ -30,6 +31,11 @@ var (
 	}
 )
 
+var (
+	// cache the first server setting when there are multiple servers
+	shortDescCache sync.Map
```
sysadmind, line 36: I think instead of adding a sync.Map, I would be more inclined to use an empty or generic static string. I don't know that caching the first and always using that value provides more value. 

Do you disagree @SuperQ?

SuperQ: Yes, I think these should be converted to static strings. In large deployments, having multiple versions of the help string bloats the `metadata` endpoints. We had similar problems with `jmx_exporter` where we had hundreds of different help strings, which made for huge (many megabyte) metadata json, which would slow down Grafana a lot.

khkhalifa: Thanks @sysadmind & @SuperQ 

I suggested a new solution. Please let me know if you expect a different approach

sysadmind:
I think this is good code-wise. The DCO check needs to be fixed before merging though. If you click the details, it will explain how to fix it. Every commit needs the DCO. It may make sense to rebase this into a single commit to make that easier.

sysadmind (approved):

PR #785, opened 2023-03-30
Add the instance struct to handle connections

sysadmind:
The intent is to use the instance struct to hold the connection to the database as well as metadata about the instance. Currently this metadata only includes the version of postgres for the instance which can be used in the collectors to decide what query to run. In the future this could hold more metadata but for now it keeps the Collector interface arguments to a reasonable number.

### DIFF
--- a/collector/instance.go
+++ b/collector/instance.go
@@ -1,0 +1,85 @@
+// Copyright 2023 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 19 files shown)


sysadmind:
@SuperQ Let me know what you think about this. This was my first iteration but it seems like it should work for versioning queries in the collector package.

SuperQ:
Seems like a good idea, care to rebase this?

sysadmind:
@SuperQ Rebase complete.

collector/instance.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return semver.Version{}, err
+	}
+	submatches := versionRegex.FindStringSubmatch(version)
+	if len(submatches) > 1 {
+		return semver.ParseTolerant(submatches[1])
+	}
+
+	// We could also try to parse the version from the server_version field.
+	// This is of the format 13.3 (Debian 13.3-1.pgdg100+1)
```
SuperQ, line 75: Maybe we should do `SHOW server_version` first?

SuperQ (approved):
Minor question about version parsing order. Otherwise LGTM.

SuperQ:
I think the version stuff is fine for now, we can iterate on it later. I'm going to merge this so I can continue some migrations of collectors out of the main package.

Issue #660, opened 2022-06-27
'postgres://' designator is not supported in connection string


Vampouille:
The exporter is not able to use connection strings that starts with `postgres://`. Only connection string starting with `postgresql://` is working.

**What did you do?**

I'm trying to connect the exporter to a managed DB

**What did you expect to see?**

Successful connection is expected as `psql` is able to connect to the DB with the same connection string:
```bash
$ psql "postgres://admin:PASSWORD_REMOVED@xxxxxxxx:21699/defaultdb?sslmode=require"
psql (14.4 (Debian 14.4-1.pgdg110+1))
Connexion SSL (protocole : TLSv1.3, chiffrement : TLS_AES_256_GCM_SHA384, bits : 256, compression : désactivé)
Saisissez « help » pour l'aide.

defaultdb=> 

```

**What did you see instead? Under which circumstances?**

An error during connection : 
```
Unable to parse DSN as either URI or connstring" dsn="postgres://admin:PASSWORD_REMOVED@xxxxxxxx:21699/defaultdb?sslmode=require
```
The [documentation](https://www.postgresql.org/docs/current/libpq-connect.html#id-1.7.3.8.3.6) explains that both designator are allowed: 

* postgresql://
* postgres://

```
The URI scheme designator can be either postgresql:// or postgres://. 
```

**Environment**

* System information:
```
Linux 5.10.0-14-amd64 x86_64
```
* postgres_exporter version:

Running as docker container : prometheuscommunity/postgres-exporter:v0.10.1
```
postgres_exporter, version 0.10.1 (branch: HEAD, revision: 6cff384d7433bcb1104efe3b496cd27c0658eb09)
  build user:       root@eb21848025d7
  build date:       20220114-17:20:30
  go version:       go1.17.6
  platform:         linux/amd64
```

* postgres_exporter flags:

No flag used but the following env vars : 

```
DATA_SOURCE_NAME=postgres://admin:PASSWORD_REMOVED@xxxxxxxx:21699/defaultdb?sslmode=require
PG_EXPORTER_AUTO_DISCOVER_DATABASES=true   
PG_EXPORTER_EXTEND_QUERY_PATH=/mnt/queries.yaml
```

* PostgreSQL version:
```
PostgreSQL 14.4 on x86_64-pc-linux-gnu, compiled by gcc, a cc67244dc0 p 47d487d188, 64-bit
```

* Logs:
```
[...]

PR #787, opened 2023-04-12
Supports alternate postgres:// prefix in URLs

mothershipper:
Adds support for the alternate postgres:// prefix in URLs. It's maybe not the cleanest approach, but works.  Hoping I can either get some pointers on a more appropriate patch, or that we could use this in the interim to unblock this use-case.

Fixes: #660

### DIFF
--- a/cmd/postgres_exporter/datasource.go
+++ b/cmd/postgres_exporter/datasource.go
@@ -35,7 +35,7 @@ func (e *Exporter) discoverDatabaseDSNs() []string {
 		var dsnURI *url.URL
 		var dsnConnstring string
 
-		if strings.HasPrefix(dsn, "postgresql://") {
+		if strings.HasPrefix(dsn, "postgresql://") || strings.HasPrefix(dsn, "postgres://") {
 			var err error
 			dsnURI, err = url.Parse(dsn)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


sysadmind:
I think this looks good. I took a look at lib/pq and this seems to be a supported scheme. @SuperQ Do you know why circleci is mad?

mothershipper:
Yeah, the happy-path (1-1 image replacement in the helm chart) works for getting our metrics exported. I haven't poked around at the entire CLI surface though, and I didn't find anything else doing checks for `postgresql://` in the repo.

I originally had the parser code overwrite the `Scheme` to be `postgresql` even when `postgres://` was supplied, but opted to revert that once I realized everything seemed ok without adjusting it.

mothershipper:
Er, regarding CircleCI -- wondering if it's because I forked the repo, neither my personal account or our company's GH account have granted Circle access before. Doesn't seem to be a way to log in via GitHub without giving them access to my private repositories which may be non-starter for me :/ 

Happy to send it over as a patch file if that turns out to be the cause.

sysadmind:
I think you could rebase on the latest master branch and circle ci should run.

sysadmind (approved):
LGTM

PR #790, opened 2023-04-22
Mixin update

v-zhuravlev:
1) [Fix cache hit ratio panel](https://github.com/prometheus-community/postgres_exporter/commit/a7197874537e0c69521ebfa3748cc01fe657b9d4)

This rounds to .001, otherwise cache hit axis was showing multiple points with 100% since 
values were rounded at axis, but not raw values,  which changes from  0.9900001 were shown 0.9900002 on graphs, causing visual noise)

2) Use psql specific metric to avoid return of postgres unrelated instances
3) [Change graphs to timeseries](https://github.com/prometheus-community/postgres_exporter/commit/add1176c0d84d42bf3c68774b0c3a4c6f26332d4)
4) Move QPS to left and remove transparency for common style
5) [Fix conflict/deadlock metrics](https://github.com/prometheus-community/postgres_exporter/commit/f342db4ffe30f4d19b9b6c003321a03466e3d74f)

These metrics are gauges, so applying rate is not required
screenshot-new

TIP: check diff by commit, for clearer view what has changed.

### DIFF
--- a/postgres_mixin/dashboards/postgres-overview.json
+++ b/postgres_mixin/dashboards/postgres-overview.json
@@ -3,79 +3,240 @@
     "list": [
       {
         "builtIn": 1,
-        "datasource": "-- Grafana --",
+        "datasource": {
+          "type": "datasource",
+          "uid": "grafana"
+        },
         "enable": true,
         "hide": true,
         "iconColor": "rgba(0, 211, 255, 1)",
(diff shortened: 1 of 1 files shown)


ArthurSens:
Hey, thanks for working on the mixins, and sorry for the super-delayed review! Looks like most of the improvements here were already added by #1179, so I'll close this one :)

PR #801, opened 2023-05-31
Move queries from queries.yaml to collectors

Sticksman:
This PR moves queries from queries.yaml into collectors. Will eventually also add additional logical replication queries.


SuperQ:
Nice, it would be useful to start adding tests. Let us know when this is ready for review.

collector/pg_replication.go
```diff
@@ -0,0 +1,60 @@
...
+				ELSE GREATEST (0, EXTRACT(EPOCH FROM (now() - pg_last_xact_replay_timestamp())))
+				END AS lag`)
+
+	var lag float64
+	err := row.Scan(&lag)
+	if err != nil {
+		return err
+	}
+	ch <- prometheus.MustNewConstMetric(
+		pgPostmaster["replication_lag"],
```
SuperQ, line 56:
```suggestion
		pgReplication["replication_lag"],
```

Sticksman:
Couldn't quite figure out how to test the histogram

collector/replication_slots.go
```diff
@@ -70,23 +73,31 @@ func (PGReplicationSlotCollector) Update(ctx context.Context, db *sql.DB, ch cha
...
+
 		if err := rows.Scan(&slot_name, &wal_lsn, &flush_lsn, &is_active); err != nil {
 			return err
 		}
 
+		if is_active {
+			is_active_value = 1
+		} else {
+			is_active_value = 0
+		}
```
SuperQ, line 86: Go style nit.

Also, I just noticed that we have a bunch of snake case in this codebase. Will have to fix that up separately.

```suggestion
		isActiveValue := 0
		if is_active {
			isActiveValue = 1
		}
```

collector/replication_slots.go
```diff
@@ -33,33 +33,36 @@ func NewPGReplicationSlotCollector(config collectorConfig) (Collector, error) {
...
-		"pg_replication_slot_confirmed_flush_lsn",
-		"last lsn confirmed flushed to the replication slot",
-		[]string{"slot_name"}, nil,
-	)
-	pgReplicationSlotIsActiveDesc = prometheus.NewDesc(
-		"pg_replication_slot_is_active",
-		"last lsn confirmed flushed to the replication slot",
-		[]string{"slot_name"}, nil,
-	)
+var pgReplicationSlotCurrentWalLSN = prometheus.NewDesc(
```
SuperQ, line 36: Maybe let's do this refactoring in a separate PR?

SuperQ (approved):
Nice

SuperQ:
@sysadmind Any comments?

brianv0:
@SuperQ @Sticksman  - So, with this PR, is queries.yaml deprecated for all use, or just for those specific queries, because the README.md now references an empty queries.yaml:

> extend.query-path Path to a YAML file containing custom queries to run. Check out [queries.yaml](https://github.com/prometheus-community/postgres_exporter/blob/master/queries.yaml) for examples of the format.

I'm sure it made sense to move some queries to their own custom collectors, but having a config-file driven definition of metric queries is incredibly useful to specify things at runtime, alternative configurations, and include metrics queries you've never thought about.

sysadmind:
We are in the process of deprecating it for all uses. The use case where custom queries are generating metrics is better served by the https://github.com/burningalchemist/sql_exporter project.

SuperQ:
@brianv0 Alternative configuraitons, queries we don't think about / know of, can be added as Go code to collectors. That's part of the reason why we're removing it. Because people customize it locally, they never contribute their improvements back to the exporter. So the whole community misses out on expertise. Requiring things be compiled in is a forcing function to make improvements be contributed.

At the same time, "user queries" / "business metrics" are _explicitly_ out of scope for this exporter. This exporter is for monitoring postgres _servers_, not the contents of the server. The linked `sql_exporter` is a better match for those use cases.

netikras:
> We are in the process of deprecating it for all uses. The use case where custom queries are generating metrics is better served by the https://github.com/burningalchemist/sql_exporter project.

Can this exact quote be added to the `queries.yaml` file as a comment explaining what's going on? Perhaps with a link to https://github.com/prometheus-community/postgres_exporter/pull/801#issuecomment-1591199421 ?

Also, update the README.md, because it still refers to the `queries.yaml` for examples, which now is empty.


To those souls googling for `queries.yaml` examples, please see [this commit](https://github.com/prometheus-community/postgres_exporter/pull/801/files#diff-996ccb161dc17b823796091b615a7a170f52ec79eaa4f69093a519498d1356f6)

prochac:
When it will be deprecated/removed? We are using this exporter for observability of our queue implementation on Postgres. The necessity of adding another exporter is very annoying.

>  That's part of the reason why we're removing it. Because people customize it locally, they never contribute their improvements back to the exporter. 
I'm not sure how others would benefit from this, LoL.

The only thing that will happen with the removal of the feature is many broken dashboards.
The possibility of custom metrics was the selling point for us.

netikras:
> This is really unfortunate, please consider reverting this useless change.

when life gives lemons...

I would never have discovered https://github.com/burningalchemist/sql_exporter hadn't this change occurred.
Now I have a database-agnostic exporter allowing for arbitrary queries :)

PR #811, opened 2023-06-13
Deprecate extend queries feature

SuperQ:
Mark the extend queries feature as deprecated in favor of recommending the sql_exporter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ This will build the docker image as `prometheuscommunity/postgres_exporter:${bra
 * `auto-discover-databases`
   Whether to discover the databases on a server dynamically.  Default is `false`.
 
-* `extend.query-path`
+* `extend.query-path` (DEPRECATED)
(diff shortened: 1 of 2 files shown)


sysadmind (approved):
Should we also mention which version we plan to remove the functionality or just add then whenever we decide?

SuperQ:
Yes, we may choose to keep it around for a bit. So let's just leave it open-ended for now.

vuducmanh11:
We are using the PostgreSQL exporter to expose some metrics of the PostgreSQL server, such as autovacuum statistics and dead/live rows. 
Because it does not support extended queries, we must deploy two exporters with the same purpose: exporting metrics of the PostgreSQL server :face_in_clouds:

SuperQ:
> We are using the PostgreSQL exporter to expose some metrics of the PostgreSQL server, such as autovacuum statistics and dead/live rows.

Are you sure there is not already a collector for this? If not, please file a new issue for feature requests like this.

PR #812, opened 2023-06-13
Fix up collector registration

SuperQ:
Use const definitions to make collector registration consistent.
* Use collector subsystem name consistently.
* Fix up replication metric name unit.


SuperQ:
Turns out there were some copy-n-paste errors in the collector names.

sysadmind (approved):

SuperSandro2000:
I think this change was missed in the changelog which is significant because some flags like ``--no-collector.bgwriter`` changed to ``--no-collector.stat_bgwriter``.

PR #815, opened 2023-06-19
Deprecate additional database features

SuperQ:
Now that we have deprecated extended queries we can deprecate related database features.
* Deprecate flags/functions around auto discover databases.
* Deprecate flags/functions for additional constant labels.


sysadmind (approved):
LGTM

PR #822, opened 2023-06-24
Add config reload via /-/reload

winfredwz:
add config.file hot reload


ArthurSens:
Hey hey, sorry for the very late review. I think this is a great feature, but since the PR was created years ago, we'd probably need to start a new one from scratch.

Just some extra comments: we might also want to accept a SIGHUP signal to trigger a hot reload, but it doesn't need to be in the same PR as the one adding the `/-/reload` route. Also, the current implementation returns 200 even for failed reloads; we probably want to return something like 500. Tests would also be nice :)

Issue #818, opened 2023-06-22
Several of the new collectors fail with incorrect data types


sclu1034:
**What did you do?**
Updated to v0.13.0 and removed `queries.yaml` in favor of the built-in collectors.

**What did you expect to see?**
An error-free log and the same metrics to be reported as before.

**What did you see instead? Under which circumstances?**

Repeating entries of the logs shown below and affected metrics, such as `pg_postmaster_start_time_seconds`, stopped being reported.
No special circumstances, I just started the tool via `quay.io/prometheuscommunity/postgres-exporter:latest` with the exact same setup as the previous image.

**Environment**

* postgres_exporter version:

```
postgres_exporter, version 0.13.0 (branch: HEAD, revision: f550f8f8ad77e054f12eb3d3f6605edc037573f5)
  build user:       root@de207ed6b751
  build date:       20230622-07:05:45
  go version:       go1.20.5
  platform:         linux/amd64
  tags:             netgo static_build
```

* postgres_exporter flags: Default Docker CMD line

* PostgreSQL version: 15.3

* Logs:
```logfmt
ts=2023-06-22T08:25:13.132Z caller=proc.go:250 msg="Excluded databases" databases=[]
ts=2023-06-22T08:25:13.142Z caller=tls_config.go:274 level=info msg="Listening on" address=[::]:9187
ts=2023-06-22T08:25:13.142Z caller=tls_config.go:277 level=info msg="TLS is disabled." http2=false address=[::]:9187
ts=2023-06-22T08:25:17.855Z caller=server.go:74 level=info msg="Established new database connection" fingerprint=postgres:5432
ts=2023-06-22T08:25:17.860Z caller=collector.go:190 level=error msg="collector failed" name=process_idle duration_seconds=0.004954557 err="sql: Scan error on column index 3, name \"seconds\": unsupported Scan, storing driver.Value type []uint8 into type *[]int64"
ts=2023-06-22T08:25:17.861Z caller=postgres_exporter.go:622 level=info msg="Semantic version changed" server=postgres:5432 from=0.0.0 to=15.3.0
ts=2023-06-22T08:25:17.864Z caller=collector.go:190 level=error msg="collector failed" name=postmaster duration_seconds=0.008552172 err="sql: Scan error on column index 0, name \"pg_postmaster_start_time\": converting driver.Value type time.Time (\"2023-06-19 13:22:47.713957 +0200 CEST\") to a float64: invalid syntax"
[...]

PR #823, opened 2023-06-24
Bug fix: Make collector not fail on null values

Sticksman:
Resolves: https://github.com/prometheus-community/postgres_exporter/issues/818

Make all row scans that gather values nullable except the ones that gather absolute values. Nulls are emitted to prometheus as "0" values or "unknown" for labels. Also wrote an additional set of tests to make sure nulls are caught and dealt with properly.

### DIFF
--- a/collector/pg_database.go
+++ b/collector/pg_database.go
@@ -106,11 +106,9 @@ func (c PGDatabaseCollector) Update(ctx context.Context, instance *instance, ch 
 			return err
 		}
 
-		var sizeMetric float64
+		sizeMetric := 0.0
 		if size.Valid {
 			sizeMetric = size.Float64
-		} else {
-			sizeMetric = 0
 		}
 		ch <- prometheus.MustNewConstMetric(
 			pgDatabaseSizeDesc,
--- a/collector/pg_database.go
+++ b/collector/pg_database.go
@@ -15,6 +15,7 @@ package collector
 
 import (
 	"context"
+	"database/sql"
 
 	"github.com/go-kit/log"
 	"github.com/prometheus/client_golang/prometheus"
@@ -79,32 +80,41 @@ func (c PGDatabaseCollector) Update(ctx context.Context, instance *instance, ch 
 	var databases []string
 
 	for rows.Next() {
-		var datname string
+		var datname sql.NullString
 		if err := rows.Scan(&datname); err != nil {
 			return err
 		}
 
+		if !datname.Valid {
+			continue
+		}
 		// Ignore excluded databases
 		// Filtering is done here instead of in the query to avoid
 		// a complicated NOT IN query with a variable number of parameters
-		if sliceContains(c.excludedDatabases, datname) {
+		if sliceContains(c.excludedDatabases, datname.String) {
 			continue
 		}
 
-		databases = append(databases, datname)
+		databases = append(databases, datname.String)
 	}
 
 	// Query the size of the databases
 	for _, datname := range databases {
-		var size int64
+		var size sql.NullFloat64
 		err = db.QueryRowContext(ctx, pgDatabaseSizeQuery, datname).Scan(&size)
 		if err != nil {
 			return err
 		}
 
+		var sizeMetric float64
+		if size.Valid {
+			sizeMetric = size.Float64
+		} else {
+			sizeMetric = 0
+		}
 		ch <- prometheus.MustNewConstMetric(
 			pgDatabaseSizeDesc,
-			prometheus.GaugeValue, float64(size), datname,
+			prometheus.GaugeValue, sizeMetric, datname,
 		)
 	}
 	if err := rows.Err(); err != nil {
--- a/collector/pg_postmaster.go
+++ b/collector/pg_postmaster.go
@@ -15,6 +15,7 @@ package collector
 
 import (
 	"context"
+	"database/sql"
 
 	"github.com/prometheus/client_golang/prometheus"
 )
@@ -51,14 +52,20 @@ func (c *PGPostmasterCollector) Update(ctx context.Context, instance *instance, 
 	row := db.QueryRowContext(ctx,
 		pgPostmasterQuery)
 
-	var startTimeSeconds float64
+	var startTimeSeconds sql.NullFloat64
 	err := row.Scan(&startTimeSeconds)
 	if err != nil {
 		return err
 	}
+	var startTimeSecondsMetric float64
+	if startTimeSeconds.Valid {
+		startTimeSecondsMetric = startTimeSeconds.Float64
+	} else {
+		startTimeSecondsMetric = 0
+	}
 	ch <- prometheus.MustNewConstMetric(
 		pgPostMasterStartTimeSeconds,
-		prometheus.GaugeValue, startTimeSeconds,
+		prometheus.GaugeValue, startTimeSecondsMetric,
 	)
 	return nil
 }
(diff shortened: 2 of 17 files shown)


collector/pg_postmaster.go
```diff
@@ -51,14 +52,20 @@ func (c *PGPostmasterCollector) Update(ctx context.Context, instance *instance,
...
 	err := row.Scan(&startTimeSeconds)
 	if err != nil {
 		return err
 	}
+	var startTimeSecondsMetric float64
+	if startTimeSeconds.Valid {
+		startTimeSecondsMetric = startTimeSeconds.Float64
+	} else {
+		startTimeSecondsMetric = 0
+	}
```
SuperQ, line 65:
```suggestion
	startTimeSecondsMetric := 0.0
	if startTimeSeconds.Valid {
		startTimeSecondsMetric = startTimeSeconds.Float64
	}
```

collector/pg_process_idle.go
```diff
@@ -97,10 +99,26 @@ func (PGProcessIdleCollector) Update(ctx context.Context, instance *instance, ch
 	if err != nil {
 		return err
 	}
+
+	var applicationNameLabel string
+	if applicationName.Valid {
+		applicationNameLabel = applicationName.String
+	} else {
+		applicationNameLabel = "unknown"
+	}
```
SuperQ, line 108:
```suggestion
	applicationNameLabel := "unknown"
	if applicationName.Valid {
		applicationNameLabel = applicationName.String
	}
```

SuperQ (commented):
Minor Go style nit.

collector/pg_database.go
```diff
@@ -79,32 +80,41 @@ func (c PGDatabaseCollector) Update(ctx context.Context, instance *instance, ch
...
 		if err != nil {
 			return err
 		}
 
+		var sizeMetric float64
+		if size.Valid {
+			sizeMetric = size.Float64
+		} else {
+			sizeMetric = 0
+		}
```
SuperQ, line 114:
```suggestion
		sizeMetric := 0.0
		if size.Valid {
			sizeMetric = size.Float64
		}
```

collector/pg_replication_slot.go
```diff
@@ -82,32 +83,44 @@ func (PGReplicationSlotCollector) Update(ctx context.Context, instance *instance
...
+		var flushLSN sql.NullFloat64
+		var isActive sql.NullBool
 		if err := rows.Scan(&slotName, &walLSN, &flushLSN, &isActive); err != nil {
 			return err
 		}
 
 		isActiveValue := 0
-		if isActive {
+		if isActive.Valid && isActive.Bool {
 			isActiveValue = 1
```
SuperQ, line 96: Can simplify to remove need for `float64()` later:

```go
		isActiveValue := 0.0
		if isActive.Valid && isActive.Bool {
			isActiveValue = 1.0
```

collector/pg_stat_database.go
```diff
@@ -280,152 +267,218 @@ func (PGStatDatabaseCollector) Update(ctx context.Context, instance *instance, c
...
+		datidLabel := "unknown"
+		if datid.Valid {
+			datidLabel = datid.String
+		}
+		datnameLabel := "unknown"
+		if datname.Valid {
+			datnameLabel = datname.String
+		}
 
+		var numBackendsMetric float64
```
SuperQ, line 279: This should probably be reset to 0 to avoid row-torow leaking int the for loop.

```suggestion
		numBackendsMetric := 0.0
```

collector/pg_stat_database_test.go
```diff
@@ -0,0 +1,220 @@
...
+			2034563757,
+			0,
+			2725688749,
+			23,
+			52,
+			74,
+			925,
+			16,
+			823,
+			srT)
```
SuperQ, line 82: Can we add a second row with all 0 entries to make sure we don't have any row-to-row data leaks?

Sticksman:
Made all `var foo float64` declarations into `foo := 0.0`. Added a row leak test, but I think it should be fine, considering lint yells at us about `var foo type` being a 0 declaration

SuperQ (approved):
Nice

collector/pg_database.go
```diff
@@ -79,32 +80,39 @@ func (c PGDatabaseCollector) Update(ctx context.Context, instance *instance, ch
 	var databases []string
 
 	for rows.Next() {
-		var datname string
+		var datname sql.NullString
```
sysadmind, line 83: Is this column nullable? The documentation doesn't mention that this field is nullable.

collector/pg_database.go
```diff
@@ -79,32 +80,39 @@ func (c PGDatabaseCollector) Update(ctx context.Context, instance *instance, ch
...
 		}
 
-		databases = append(databases, datname)
+		databases = append(databases, datname.String)
 	}
 
 	// Query the size of the databases
 	for _, datname := range databases {
-		var size int64
+		var size sql.NullFloat64
```
sysadmind, line 103: I also don't see any documentation that says that this field can be NULL>

sysadmind (changes requested):
The more that I look at this PR, the more that it appears it is blindly using nullable data types for every query. I don't think that's the best approach. I think that the data types should change only where the field can actually be null.

SuperQ:
@sysadmind I think we should handle nullable for basically everything, since you never know when PG will either change the underlying data between versions, or have incorrect documentation. :-/

Sticksman:
datname shouldn't be nullable, but one the original report included it as a field that returned null: https://github.com/prometheus-community/postgres_exporter/issues/818#issuecomment-1604358077

collector/pg_process_idle.go
```diff
@@ -15,21 +15,23 @@ package collector
...
 	"github.com/go-kit/log"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-const processIdleSubsystem = "process_idle"
-
 func init() {
-	registerCollector(processIdleSubsystem, defaultEnabled, NewPGProcessIdleCollector)
+	// Making this default disabled because we have no tests for it
+	registerCollector("statements", defaultDisabled, NewPGProcessIdleCollector)
```
sysadmind, line 26: Why was this renamed?

sysadmind (commented):
One question this round. 

@SuperQ I'm not sure that I agree that everything should be NULLable. This adds boilerplate for every field and makes the code harder to read. If you feel that this is the best path, I won't protest but I don't love it.

SuperQ:
I don't love it either, but I don't know enough about PostgreSQL to know if we should one-off fix NULL errors or deal with the boilerplate.

collector/pg_process_idle.go
Sticksman: Good question. Probably a typo somewhere

collector/pg_stat_database_test.go
sysadmind: This shows as resolved, but I don't see any of these with 2 rows

SuperQ: Oh, I saw some second rows in other tests.

SuperQ: Oh, yea, see `TestPGStatDatabaseCollectorRowLeakTest` below.

sysadmind (approved):
I don't think that this is as clean as I would like but I think it is at a state that we can iterate from and at least resolve the bugs for our users.

SuperQ (approved):

PR #826, opened 2023-06-27
Cleanup collectors

SuperQ:
Fix up `replication` and `process_idle` Update input params to match the rest of the collectors.

### DIFF
--- a/collector/pg_replication.go
+++ b/collector/pg_replication.go
@@ -15,7 +15,6 @@ package collector
 
 import (
 	"context"
-	"database/sql"
 
 	"github.com/prometheus/client_golang/prometheus"
 )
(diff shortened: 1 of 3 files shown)


collector/pg_replication.go
```diff
@@ -64,7 +63,8 @@ var (
 	END as is_replica`
 )
 
-func (c *PGReplicationCollector) Update(ctx context.Context, db *sql.DB, ch chan<- prometheus.Metric) error {
```
sysadmind, line 67: weird, I thought that we fixed this in another PR.

sysadmind (approved):
LGTM. Thanks

collector/pg_replication.go
SuperQ: I think this was a separate collector that got merged midway.

PR #819, opened 2023-06-23
Cleanup/add gitlab exporters

Sticksman:
This adds most of the diagonistic queries found in the gitlab queries.yaml: https://gitlab.com/gitlab-cookbooks/gitlab-exporters/-/blob/master/templates/postgres_exporter/queries.yaml.erb#L866

Several are difficult to write tests for:
- pg_xid has an Nan comparison that can't be made, so had to special case that test
- the summary queries use some pretty complex regexes that are difficult to sanitize, which means tests were nearly impossible to write. I checked on a postgres db and they seemed to work alright.

PR #835, opened 2023-06-28
Gitlab Collector: Index size collector and test

Sticksman:
Split from #819

### DIFF
--- a/collector/pg_index_size.go
+++ b/collector/pg_index_size.go
@@ -1,0 +1,106 @@
+// Copyright 2023 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package collector
+
+import (
+	"context"
+	"database/sql"
+
+	"github.com/go-kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const indexSizeSubsystem = "index_size"
+
+func init() {
+	registerCollector(indexSizeSubsystem, defaultDisabled, NewPGIndexSizeCollector)
+}
+
+type PGIndexSizeCollector struct {
+	log log.Logger
+}
+
+func NewPGIndexSizeCollector(config collectorConfig) (Collector, error) {
+	return &PGIndexSizeCollector{log: config.logger}, nil
+}
+
+var (
+	indexSizeDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, indexSizeSubsystem, "bytes"),
+		"Size of the index as per pg_table_size function",
+		[]string{"schemaname", "relname", "indexrelname"},
+		prometheus.Labels{},
+	)
+
+	indexSizeQuery = `
+	SELECT
+		schemaname,
+		tablename as relname,
+		indexname as indexrelname,
+		pg_class.relpages * 8192::bigint as index_size
+	FROM
+		pg_indexes inner join pg_namespace on pg_indexes.schemaname = pg_namespace.nspname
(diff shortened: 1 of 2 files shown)


collector/pg_index_size.go
```diff
@@ -0,0 +1,106 @@
...
+	SELECT
+		schemaname,
+		tablename as relname,
+		indexname as indexrelname,
+		pg_class.relpages * 8192::bigint as index_size
+	FROM
+		pg_indexes inner join pg_namespace on pg_indexes.schemaname = pg_namespace.nspname
+		inner join pg_class on pg_class.relnamespace = pg_namespace.oid and pg_class.relname = pg_indexes.indexname
+	WHERE
+		pg_indexes.schemaname != 'pg_catalog'
```
sysadmind, line 56:
```suggestion
	FROM
		pg_indexes 
		INNER JOIN pg_namespace 
			ON pg_indexes.schemaname = pg_namespace.nspname
		INNER JOIN pg_class 
			ON pg_class.relnamespace = pg_namespace.oid AND pg_class.relname = pg_indexes.indexname
	WHERE
		pg_indexes.schemaname != 'pg_catalog'
```

collector/pg_index_size.go
```diff
@@ -0,0 +1,106 @@
...
+		[]string{"schemaname", "relname", "indexrelname"},
+		prometheus.Labels{},
+	)
+
+	indexSizeQuery = `
+	SELECT
+		schemaname,
+		tablename as relname,
+		indexname as indexrelname,
+		pg_class.relpages * 8192::bigint as index_size
```
sysadmind, line 51:
```suggestion
		tablename AS relname,
		indexname AS indexrelname,
		pg_class.relpages * 8192::bigint AS index_size
```

collector/pg_index_size.go
```diff
@@ -0,0 +1,106 @@
...
+	defer rows.Close()
+
+	for rows.Next() {
+		var schemaname, relname, indexrelname sql.NullString
+		var indexSize sql.NullFloat64
+
+		if err := rows.Scan(&schemaname, &relname, &indexrelname, &indexSize); err != nil {
+			return err
+		}
+		schemanameLabel := "unknown"
```
sysadmind, line 77: I'm really not sure that these metrics even make sense if all of these are NULL. I understand that we used NULL to fix all of the bugs in the 0.13.1 release, but these are net new collectors and I would like to be much more thoughtful about how we handle them in code. For example, if the indexSize metric is NULL, why would we even report the metric? How is the zero value useful.

sysadmind (changes requested):
Thank you for your work on this collector and the other PRs that you have submitted. I know that it took a lot of time to put these together and I appreciate it. I have a few comments on this PR. I will try to get to the other PRs soon, but I think much of the feedback will be similar. My priority in reviewing the code is to consider how we can maintain this project long term. I am thrilled with the forward progress on this exporter but I don't want to sacrifice quality for the sake along the way. 

As for the SQL formatting, I am particular about this. Poorly formatted SQL is much harder to read and understand. We do not have a style guide and I haven't seen a good formatter for SQL inside of go yet so I don't have anything good to help you directly. This site looks like it is close to what I would like to see: https://www.sqlstyle.guide/ but I haven't read it in detail yet. I think as long as we are close with style, we can make it work.

collector/pg_index_size.go
SuperQ: Agreed, if the labels are null, we should Debug log and `continue`.

Sticksman: Makes sense to me

sysadmind: I think this code is still emitting metrics when these values are NULL.

collector/pg_index_size.go
```diff
@@ -0,0 +1,109 @@
...
+		[]string{"schemaname", "relname", "indexrelname"},
+		prometheus.Labels{},
+	)
+
+	indexSizeQuery = `
+	SELECT
+		schemaname,
+		tablename AS relname,
+		indexname AS indexrelname,
+		pg_class.relpages * 8192::bigint AS index_size
```
sysadmind, line 51: We can't rely on 8192 being the block size. It's default, but could be changed. See references:
1. https://www.postgresql.org/docs/current/catalog-pg-class.html (see relpages)
2. https://www.postgresql.org/docs/current/runtime-config-preset.html (see block_size)

sysadmind (changes requested):

collector/pg_index_size.go
Sticksman: ok so we have to select block size and multiply by that

PR #836, opened 2023-06-28
Gitlab Collector: Long running transactions collector and test

Sticksman:
Split from #819

### DIFF
--- a/collector/pg_long_running_transactions.go
+++ b/collector/pg_long_running_transactions.go
@@ -43,7 +43,7 @@ var (
 	)
 
 	longRunningTransactionsAgeInSeconds = prometheus.NewDesc(
-		prometheus.BuildFQName(namespace, longRunningTransactionsSubsystem, "age_in_seconds"),
+		prometheus.BuildFQName(namespace, longRunningTransactionsSubsystem, "oldest_timestamp_seconds"),
 		"The current maximum transaction age in seconds",
 		[]string{},
 		prometheus.Labels{},
--- a/collector/pg_long_running_transactions.go
+++ b/collector/pg_long_running_transactions.go
@@ -54,7 +54,7 @@ var (
 		COUNT(*) as transactions,
    		MAX(EXTRACT(EPOCH FROM clock_timestamp())) AS oldest_timestamp_seconds
     FROM pg_catalog.pg_stat_activity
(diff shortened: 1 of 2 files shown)


collector/pg_long_running_transactions.go
```diff
@@ -0,0 +1,93 @@
...
+var (
+	longRunningTransactionsCount = prometheus.NewDesc(
+		"pg_long_running_transactions",
+		"Current number of long running transactions",
+		[]string{},
+		prometheus.Labels{},
+	)
+
+	longRunningTransactionsAgeInSeconds = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, longRunningTransactionsSubsystem, "age_in_seconds"),
```
SuperQ, line 46:
```suggestion
		prometheus.BuildFQName(namespace, longRunningTransactionsSubsystem, "oldest_timestamp_seconds"),
```

collector/pg_long_running_transactions.go
```diff
@@ -0,0 +1,93 @@
...
+		"The current maximum transaction age in seconds",
+		[]string{},
+		prometheus.Labels{},
+	)
+
+	longRunningTransactionsQuery = `
+	SELECT
+		COUNT(*) as transactions,
+   		MAX(EXTRACT(EPOCH FROM (clock_timestamp() - xact_start))) AS age_in_seconds
```
SuperQ, line 55:
```suggestion
   		MAX(EXTRACT(EPOCH FROM clock_timestamp())) AS oldest_timestamp_seconds
```

collector/pg_long_running_transactions.go
```diff
@@ -0,0 +1,93 @@
...
+    FROM pg_catalog.pg_stat_activity
+    WHERE state is distinct from 'idle' AND (now() - xact_start) > '1 minutes'::interval AND query not like 'autovacuum:%'
```
SuperQ, line 57: I wonder if we shouldn't remove the `now() - xact_start` bit.

```suggestion
    WHERE state is distinct from 'idle' AND query not like 'autovacuum:%'
```

SuperQ (changes requested):

collector/pg_long_running_transactions.go
sberlin: How is this returning the oldest timestamp? Please take a look at #978

PR #838, opened 2023-06-28
Gitlab Collector: Oldest Blocked collector and test

Sticksman:
Split from #819

### DIFF
--- a/collector/pg_oldest_blocked.go
+++ b/collector/pg_oldest_blocked.go
@@ -1,0 +1,83 @@
+// Copyright 2023 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


collector/pg_oldest_blocked.go
```diff
@@ -0,0 +1,83 @@
...
+	oldestBlockedAgeSeconds = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, oldestBlockedSubsystem, "age_seconds"),
+		"Largest number of seconds any transaction is currently waiting on a lock",
+		[]string{},
+		prometheus.Labels{},
+	)
+
+	oldestBlockedQuery = `
+	SELECT
+		coalesce(extract('epoch' from max(clock_timestamp() - state_change)), 0) age_seconds
```
SuperQ, line 47: Similar to others, we should extract the raw timestamp for this metric.

SuperQ (changes requested):
I think we should refactor these separate collectors into a single `pg_stat_activity` collector to reduce the number of queries to this table.

collector/pg_oldest_blocked.go
Sticksman: I don't know if that makes sense since we're trying to measure the max duration that a query's been blocked here

SuperQ: This can still be done with `time() - metric`.

PR #840, opened 2023-06-28
Gitlab Collector: Autovacuum collector and test

Sticksman:
Split from #819


collector/pg_stat_activity_autovacuum.go
```diff
@@ -0,0 +1,85 @@
...
+	log log.Logger
+}
+
+func NewPGStatActivityAutovacuumCollector(config collectorConfig) (Collector, error) {
+	return &PGStatActivityAutovacuumCollector{log: config.logger}, nil
+}
+
+var (
+	statActivityAutovacuumAgeInSeconds = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, statActivityAutovacuumSubsystem, "age_in_seconds"),
```
sysadmind, line 39: @SuperQ should this be "age_seconds"?

collector/pg_stat_activity_autovacuum.go
```diff
@@ -0,0 +1,85 @@
...
+	)
+
+	statActivityAutovacuumQuery = `
+    SELECT
+		SPLIT_PART(query, '.', 2) AS relname,
+		EXTRACT(EPOCH FROM (clock_timestamp() - xact_start)) AS age_in_seconds
+    FROM
+    	pg_catalog.pg_stat_activity
+    WHERE
+		query like 'autovacuum:%' AND
```
sysadmind, line 52:
```suggestion
			query like 'autovacuum:%' AND
```

collector/pg_stat_activity_autovacuum.go
```diff
@@ -0,0 +1,85 @@
...
+		prometheus.BuildFQName(namespace, statActivityAutovacuumSubsystem, "age_in_seconds"),
+		"The age of the vacuum process in seconds",
+		[]string{"relname"},
+		prometheus.Labels{},
+	)
+
+	statActivityAutovacuumQuery = `
+    SELECT
+		SPLIT_PART(query, '.', 2) AS relname,
+		EXTRACT(EPOCH FROM (clock_timestamp() - xact_start)) AS age_in_seconds
```
sysadmind, line 48:
```suggestion
			SPLIT_PART(query, '.', 2) AS relname,
			EXTRACT(EPOCH FROM (clock_timestamp() - xact_start)) AS age_in_seconds
```

sysadmind: I think that we can use xact_start by itself without the other calculations. We should be able to use a time.Time for it. The use of clock_timestamp is the only thing making me question if this current query isn't better though. Otherwise, it would be best to also query the clock_timestamp as another field and calulate ourselves.

collector/pg_stat_activity_autovacuum.go
```diff
@@ -0,0 +1,85 @@
...
+	statActivityAutovacuumAgeInSeconds = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, statActivityAutovacuumSubsystem, "age_in_seconds"),
+		"The age of the vacuum process in seconds",
+		[]string{"relname"},
+		prometheus.Labels{},
+	)
+
+	statActivityAutovacuumQuery = `
+    SELECT
+		SPLIT_PART(query, '.', 2) AS relname,
```
sysadmind, line 47: This seems like it could be fragile. How do we guarantee that this will be the relname?

collector/pg_stat_activity_autovacuum.go
SuperQ: Yes, we should convert these from `age_seconds` to `timestamp_seconds` to better fit with best practices.

collector/pg_stat_activity_autovacuum.go
Sticksman: I guess because since the query is autogenerated it'll always look something like `autovacuum: COMMAND schema.relname`

SuperQ (approved):

collector/pg_stat_activity_autovacuum.go
```diff
@@ -0,0 +1,85 @@
...
+		EXTRACT(xact_start) AS timestamp_seconds
+    FROM
+    	pg_catalog.pg_stat_activity
+    WHERE
+		query LIKE 'autovacuum:%' AND
+		EXTRACT(EPOCH FROM (clock_timestamp() - xact_start)) > 300
```
SuperQ, line 53: I wonder if we should drop this filter and always show active autovacuums. This way users can chose what to do with the age in PromQL.

sysadmind (approved):

PR #841, opened 2023-06-28
Gitlab Collector: Autovacuum active collector and test

Sticksman:
Also add some more escapes to make tests work
Split from #819

### DIFF
--- a/collector/pg_stat_activity_autovacuum_active.go
+++ b/collector/pg_stat_activity_autovacuum_active.go
@@ -46,21 +46,12 @@ var (
 	SELECT
 		v.phase,
 		CASE
-<<<<<<< HEAD
-			when a.query ~ '^autovacuum.*to prevent wraparound' then 'wraparound'
-			when a.query ~* '^vacuum' then 'user'
-			when a.pid is null then 'idle'
-		ELSE 'regular'
-		END AS mode,
-		count(1) AS workers_count
-=======
 			WHEN a.query ~ '^autovacuum.*to prevent wraparound' THEN 'wraparound'
 			WHEN a.query ~* '^vacuum' THEN 'user'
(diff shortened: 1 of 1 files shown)


collector/pg_stat_activity_autovacuum_active.go
```diff
@@ -0,0 +1,91 @@
...
+
+	statActivityAutovacuumActiveQuery = `
+	SELECT
+		v.phase,
+		CASE
+			when a.query ~ '^autovacuum.*to prevent wraparound' then 'wraparound'
+			when a.query ~* '^vacuum' then 'user'
+			when a.pid is null then 'idle'
+		ELSE 'regular'
+		END as mode,
```
sysadmind, line 53:
```suggestion
		END AS mode,
```

collector/pg_stat_activity_autovacuum_active.go
```diff
@@ -0,0 +1,91 @@
...
+		count(1) as workers_count
```
sysadmind, line 54:
```suggestion
		count(1) AS workers_count
```

collector/pg_stat_activity_autovacuum_active.go
```diff
@@ -0,0 +1,91 @@
...
+	rows, err := db.QueryContext(ctx,
+		statActivityAutovacuumActiveQuery)
+
+	if err != nil {
+		return err
+	}
+	defer rows.Close()
+
+	for rows.Next() {
+		var phase, mode string
```
sysadmind, line 72: It would seem that the phase label would change often during vacuum, correct? That would change this time series often.

Sticksman: It doesn't change terribly often, depends on the size of the table

PR #843, opened 2023-06-28
Gitlab Collector: User Indexes collector and test

Sticksman:
Split from #819

### DIFF
--- a/collector/pg_stat_user_indexes.go
+++ b/collector/pg_stat_user_indexes.go
@@ -1,0 +1,136 @@
+// Copyright 2023 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package collector
+
+import (
+	"context"
+	"database/sql"
+
+	"github.com/go-kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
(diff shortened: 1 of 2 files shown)


collector/pg_stat_user_indexes.go
```diff
@@ -0,0 +1,136 @@
...
+	db := instance.getDB()
+	rows, err := db.QueryContext(ctx,
+		statUserIndexesQuery)
+
+	if err != nil {
+		return err
+	}
+	defer rows.Close()
+	for rows.Next() {
+		var schemaname, relname, indexrelname sql.NullString
```
sysadmind, line 79: Again here, I don't know how useful these metrics are if we have NULL returned. If the value is NULL, the metrics are potentially worse than meaningless, but rather misleading. If we decide that these need to stay NULLable, I would argue that at minimum we should skip rows where there are NULL values.

collector/pg_stat_user_indexes_test.go
```diff
@@ -0,0 +1,113 @@
...
+	ch := make(chan prometheus.Metric)
+	go func() {
+		defer close(ch)
+		c := PGStatUserIndexesCollector{}
+
+		if err := c.Update(context.Background(), inst, ch); err != nil {
+			t.Errorf("Error calling PGStatUserIndexesCollector.Update: %s", err)
+		}
+	}()
+	expected := []MetricResult{
```
sysadmind, line 99: This tests shows how the NULL values could be misleading. Not only will this take up space in the time series database, but it also provides negative value.

collector/pg_stat_user_indexes.go
Sticksman: If that's the case, perhaps we should just continue if we're missing schema, rel, or index rel?

collector/pg_stat_user_indexes_test.go
Sticksman: Do you have a recommendation? Fail on nulls?

Sticksman: Gonna go with continues on null with debug logs

SuperQ (approved):

PR #844, opened 2023-06-28
Gitlab Collector: Wal Receiver Collector and Test

Sticksman:
Split from #819 

Will have to do a merge on collector test with #841


collector/pg_stat_walreceiver.go
```diff
@@ -0,0 +1,306 @@
...
+	WHERE
+		table_name = 'pg_stat_wal_receiver' and
+		column_name = 'flushed_lsn'
+	`
+
+	pgStatWalReceiverQueryWithNoFlushedLSN = `
+	SELECT
+		trim(both '''' from substring(conninfo from 'host=([^ ]*)')) as upstream_host,
+		slot_name,
+		case status
```
sysadmind, line 112: I'm not sure it makes sense to use such a CASE statement when the query is defined in code. We can do the translation here in the exporter in code. Not using CASE makes the query much easier to read.

collector/pg_stat_walreceiver.go
```diff
@@ -0,0 +1,306 @@
...
+			when 'stopped' then 0
+			when 'starting' then 1
+			when 'streaming' then 2
+			when 'waiting' then 3
+			when 'restarting' then 4
+			when 'stopping' then 5 else -1
+		end as status,
+		(receive_start_lsn- '0/0') % (2^52)::bigint as receive_start_lsn,
+		receive_start_tli,
+		(flushed_lsn- '0/0') % (2^52)::bigint as flushed_lsn,
```
sysadmind, line 145: If this is the only field that is different between the two queries, I would prefer to use something like fmt.Sprintf to stick this into a single query. There's a lot here and any future changes would need to be careful to make the same change in both queries.

collector/pg_stat_walreceiver.go
```diff
@@ -0,0 +1,306 @@
...
+		query = pgStatWalReceiverQueryWithFlushedLSN
+	} else {
+		query = pgStatWalReceiverQueryWithNoFlushedLSN
+	}
+	rows, err := db.QueryContext(ctx, query)
+	if err != nil {
+		return err
+	}
+	defer rows.Close()
+	for rows.Next() {
```
sysadmind, line 176: I don't think that the NULL values are useful to report as metrics.

collector/pg_stat_walreceiver.go
```diff
@@ -0,0 +1,306 @@
...
+			level.Debug(c.log).Log("msg", "Skipping wal receiver stats because slotname host is null")
+			continue
+		}
+		labels := []string{upstreamHost.String, slotName.String}
+		if !status.Valid {
+			level.Debug(c.log).Log("msg", "Skipping wal receiver stats because status is null")
+			continue
+		}
+
+		var statusMetric float64
```
SuperQ, line 176: We don't do value-based status in Prometheus. The standard practice for this is to have these values be a label on an info metric with a constant value of `1`.

collector/pg_stat_walreceiver.go
```diff
@@ -0,0 +1,306 @@
...
+	return &PGStatWalReceiverCollector{log: config.logger}, nil
+}
+
+var (
+	statWalReceiverStatus = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, statWalReceiverSubsystem, "status"),
+		"Activity status of the WAL receiver process",
+		[]string{"upstream_host", "slot_name"},
+		prometheus.Labels{},
+	)
```
SuperQ, line 45:
```suggestion
	statWalReceiverStatus = prometheus.NewDesc(
		prometheus.BuildFQName(namespace, statWalReceiverSubsystem, "status_info"),
		"Activity status of the WAL receiver process",
		[]string{"upstream_host", "slot_name", "status"},
		prometheus.Labels{},
	)
```

SuperQ (changes requested):

collector/pg_stat_walreceiver.go
```diff
@@ -0,0 +1,277 @@
...
+			statWalReceiverLatestEndLsn,
+			prometheus.CounterValue,
+			latestEndLsnMetric,
+			labels...)
+
+		latestEndTimeMetric := float64(latestEndTime.Float64)
+		ch <- prometheus.MustNewConstMetric(
+			statWalReceiverLatestEndTime,
+			prometheus.CounterValue,
+			latestEndTimeMetric,
```
SuperQ, line 263: Minor nit, we don't need to do these assigns.

```suggestion
		ch <- prometheus.MustNewConstMetric(
			statWalReceiverLatestEndTime,
			prometheus.CounterValue,
			latestEndTime.Float64,
```

Sticksman:
Yeah I think it made a lot more sense when we were actually checking for null values, but now we can probably get rid of them

Sticksman:
Probably makes sense to do a cleanup PR when I'm done

collector/pg_stat_walreceiver.go
```diff
@@ -0,0 +1,268 @@
...
+		column_name = 'flushed_lsn'
+	`
+
+	pgStatWalReceiverQueryTemplate = `
+	SELECT
+		trim(both '''' from substring(conninfo from 'host=([^ ]*)')) as upstream_host,
+		slot_name,
+		status,
+		(receive_start_lsn- '0/0') % (2^52)::bigint as receive_start_lsn,
+		%sreceive_start_tli,
```
sysadmind, line 111:
```suggestion
		%s
receive_start_tli,
```
Not having whitespace here makes it hard to see this template variable. I think it's more important for the query to look easy to read here than it is after templating.

sysadmind (approved):
Just one formatting request.

dswarbrick:
Tests fail on 32-bit platforms.
```
# github.com/prometheus-community/postgres_exporter/collector [github.com/prometheus-community/postgres_exporter/collector.test]
collector/pg_stat_walreceiver_test.go:67:4: cannot use 1200668684563608 (untyped int constant) as int value in argument to sqlmock.NewRows(columns).AddRow (overflows)
collector/pg_stat_walreceiver_test.go:69:4: cannot use 1200668684563609 (untyped int constant) as int value in argument to sqlmock.NewRows(columns).AddRow (overflows)
collector/pg_stat_walreceiver_test.go:73:4: cannot use 1200668684563610 (untyped int constant) as int value in argument to sqlmock.NewRows(columns).AddRow (overflows)
collector/pg_stat_walreceiver_test.go:146:4: cannot use 1200668684563608 (untyped int constant) as int value in argument to sqlmock.NewRows(columns).AddRow (overflows)
collector/pg_stat_walreceiver_test.go:151:4: cannot use 1200668684563610 (untyped int constant) as int value in argument to sqlmock.NewRows(columns).AddRow (overflows)
ok      github.com/prometheus-community/postgres_exporter/cmd/postgres_exporter 0.007s
FAIL    github.com/prometheus-community/postgres_exporter/collector [build failed]
ok      github.com/prometheus-community/postgres_exporter/config        0.005s
FAIL
```
This is yet another occasion where adding 32-bit testing (e.g. `GOARCH=386`) to CI would have caught this, a la #857.

PR #846, opened 2023-06-28
Gitlab Collector: Stuck idle in transaction collector and test

Sticksman:
Split from #819


SuperQ (approved):

tomhughes:
Is hard coding a specific time limit in the collector really a good idea?

We've been using a [modified version](https://github.com/openstreetmap/chef/blob/90e757a0c0f3daaabc6da84c8f1c196ce2e27e2b/cookbooks/postgresql/templates/default/postgres_queries.yml.erb#L1) of the pg_process_idle query which includes both `idle` and `idle in transaction` and breaks the metrics down by both state and application which then allowed us to write alerts on `pg_process_idle_seconds` which look for [excessive delays](https://github.com/openstreetmap/chef/blob/90e757a0c0f3daaabc6da84c8f1c196ce2e27e2b/cookbooks/prometheus/templates/default/alert_rules.yml.erb#L538).

SuperQ:
@tomhughes Yea, you're correct, it would be good to have a flag for this.

Perhaps you would like to contribute your version as a new collector method?

SuperQ:
@tomhughes Another thing I was thinking about was if we wrote a `pg_stat_activity` collector that provided several different metrics. Possibly dump a raw output of the table and process it in Go to produce various metrics rather than try and do everything in SQL.

Sticksman:
Should I close this one then?

SuperQ:
@Sticksman Yes, I think the other metric is a superset of this. Do you agree?

Sticksman:
Yep, sounds good. I'll close this one

PR #847, opened 2023-06-28
Gitlab collector: Total relation size collector and test

Sticksman:
Split from #819


collector/pg_total_relation_size.go
```diff
@@ -0,0 +1,98 @@
...
+		[]string{"schemaname", "relname"},
+		prometheus.Labels{},
+	)
+
+	totalRelationSizeQuery = `
+		SELECT
+			relnamespace::regnamespace as schemaname,
+			relname as relname,
+			pg_total_relation_size(oid) bytes
+		FROM pg_class
```
SuperQ, line 51: It seems like there are several collectors that query `pg_class`. I wonder if we should refactor this into a `pg_class` collector that gathers all the data in one shot.

PR #848, opened 2023-06-28
Gitlab Collector: Xid collector and test

Sticksman:
Split from #819


collector/pg_xid.go
```diff
@@ -0,0 +1,99 @@
...
+	log log.Logger
+}
+
+func NewPGXidCollector(config collectorConfig) (Collector, error) {
+	return &PGXidCollector{log: config.logger}, nil
+}
+
+var (
+	xidCurrent = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, xidSubsystem, "current"),
```
SuperQ, line 39: Are we sure this is a monotonic counter?

PR #849, opened 2023-06-28
Gitlab collector: Xlog location collector and test

Sticksman:
Split from #819

### DIFF
--- a/collector/pg_xlog_location.go
+++ b/collector/pg_xlog_location.go
@@ -1,0 +1,80 @@
+// Copyright 2023 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


collector/pg_xlog_location.go
```diff
@@ -0,0 +1,80 @@
...
+	for rows.Next() {
+		var bytes float64
+
+		if err := rows.Scan(&bytes); err != nil {
+			return err
+		}
+
+		ch <- prometheus.MustNewConstMetric(
+			xlogLocationBytes,
+			prometheus.CounterValue,
```
SuperQ, line 72: Are we sure this is a counter? I'm not sure the data is monotonic.

Sticksman: Yeah it's not. I'll change it to Gauge

SuperQ (approved):

sysadmind (approved):

PR #854, opened 2023-07-03
Fix replication collector

tomhughes:
Make `NewPGReplicationCollector` actually construct a replication collector, not a postmaster collector.


sysadmind (approved):
Nice catch. Thanks!

PR #855, opened 2023-07-04
Unpack postgres arrays for process idle times correctly

tomhughes:
The previous code just errored calling `Scan` on the row as the arrays are returned as strings that need to be decoded with `pq.Array`.

### DIFF
--- a/collector/pg_process_idle.go
+++ b/collector/pg_process_idle.go
@@ -83,7 +83,7 @@ func (PGProcessIdleCollector) Update(ctx context.Context, instance *instance, ch
 			GROUP BY 1, 2, 3;`)
 
 	var applicationName sql.NullString
-	var secondsSum sql.NullInt64
+	var secondsSum sql.NullFloat64
 	var secondsCount sql.NullInt64
 	var seconds []float64
 	var secondsBucket []int64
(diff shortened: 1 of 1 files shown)


tomhughes:
Add a second commit to fix the type of the total seconds, which should be a float.

collector/pg_process_idle.go
```diff
@@ -111,7 +124,7 @@ func (PGProcessIdleCollector) Update(ctx context.Context, instance *instance, ch
 	}
 	secondsSumMetric := 0.0
 	if secondsSum.Valid {
-		secondsSumMetric = float64(secondsSum.Int64)
+		secondsSumMetric = float64(secondsSum.Float64)
```
SuperQ, line 127: Simplification?

```suggestion
		secondsSumMetric = secondsSum.Float64
```

SuperQ (commented):
Minor nit

collector/pg_process_idle.go
tomhughes: I don't know go at all, never mind the library that is using but if `Float64` is just a normal float that doesn't need a cast then sure.

tomhughes: Seems to work so I've made the change.

sysadmind (approved):
I think we could just change the var to a pq.Array instead of manually scanning afterwards, but I think that this is better than nothing.

tomhughes:
Oh I didn't realise that was possible... I was basing it on what the old queries based code was doing!

SuperQ:
@tomhughes Want to make any changes or should we merge this as-is?

tomhughes:
Sorry, was out last night so wasn't able to look at that suggestion... I've made the change to bind the arrays directly now.

SuperQ:
Nice

Issue #850, opened 2023-06-28
 to a float64: invalid syntax


Alexufo:
v0.13.1
pg: v15
ubuntu

In log:
pg_exporter       | ts=2023-06-28T21:12:09.788Z caller=collector.go:190 level=error msg="collector failed" name=replication duration_seconds=0.000349425 err="sql: Scan error on column index 0, name \"pg_postmaster_start_time\": converting driver.Value type time.Time (\"2023-06-27 21:20:59.88301 +0000 UTC\") to a float64: invalid syntax"
pg_exporter       | ts=2023-06-28T21:12:09.795Z caller=collector.go:190 level=error msg="collector failed" name=postmaster duration_seconds=0.007301787 err="sql: Scan error on column index 0, name \"pg_postmaster_start_time\": converting driver.Value type time.Time (\"2023-06-27 21:20:59.88301 +0000 UTC\") to a float64: invalid syntax"

baerwang:
logger time output and error time inconsistency

<img width="2443" alt="image" src="https://github.com/prometheus-community/postgres_exporter/assets/52104949/9577c6fc-4143-41c4-9f94-25bcb78885e8">

sysadmind:
This was fixed in #828 and will be part of the next release.

Issue #830, opened 2023-06-27
Improve pg_postmaster collector


sysadmind:
## Proposal
**Use case. Why is this important?**
The[ pg_postmaster collector](https://github.com/prometheus-community/postgres_exporter/blob/030a2a9bc713097e93d037b6c7031114e21dbd60/collector/pg_postmaster.go) was ported from the queries.yaml file to a builtin collector in #801. This collector may need a larger discussion about how it should work. This issue is to track those changes and discussions.

PR #828, opened 2023-06-27
Bug Fix: Fix lingering type issues

Sticksman:
Several random fixes:

1. pg_postmaster_start_time() returns a date which turns into a string, we need to convert it into a unix timestamp (or use time.time in the collector and convert to float64)
2. Sounds like process_idle is using uint for whatever reason?

Resolves #818

PR #865, opened 2023-07-07
added decimal point truncate to postmaster start time query

tschuyebuhl:
```
postgres=# SELECT CAST(extract(epoch from pg_postmaster_start_time()) AS INTEGER) as start_time;
 start_time 
------------
 1667939963
(1 row)

postgres=# SELECT extract(epoch from pg_postmaster_start_time()) as start_time;
    start_time     
-------------------
 1667939963.007028
(1 row)
```

the current query results in this error spam:

`ts=2023-07-07T15:35:20.847Z caller=collector.go:190 level=error msg="collector failed" name=postmaster duration_seconds=0.016613707 err="sql: Scan error on column index 0, name \"pg_postmaster_start_time\": converting driver.Value type time.Time (\"2022-11-08 21:39:23.007028 +0100 CET\") to a float64: invalid syntax"`
we probably dont need to use float64 with this change, but i didn't want to make a bigger mess for now
issue ref: https://github.com/prometheus-community/postgres_exporter/issues/850

### DIFF
--- a/collector/pg_postmaster.go
+++ b/collector/pg_postmaster.go
@@ -44,7 +44,7 @@ var (
 		[]string{}, nil,
 	)
 
-	pgPostmasterQuery = "SELECT extract(epoch from pg_postmaster_start_time) from pg_postmaster_start_time();"
+	pgPostmasterQuery = "SELECT CAST(extract(epoch from pg_postmaster_start_time()) AS INTEGER) as start_time;" //truncates the decimal part
 )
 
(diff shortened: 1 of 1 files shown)


sysadmind:
I just checked locally, this has already been resolved in the master branch. v0.13.1 still has this bug. The error you see was resolved in #828 

Also of note, we have opened #830 to track improvements to this collector.

tschuyebuhl:
Right. We can close this one. I just double checked now and, of course, you are right:
image
Thank you for your time anyways. Waiting til #828 makes its way downstream.

PR #877, opened 2023-07-27
Bugfix: Make statsreset nullable

Sticksman:
Null is a legit value for statsReset and we should treat it as such. That being said, not clear what the value should be if the timestamp is null

### DIFF
--- a/collector/pg_stat_database.go
+++ b/collector/pg_stat_database.go
@@ -348,8 +348,7 @@ func (c *PGStatDatabaseCollector) Update(ctx context.Context, instance *instance
 		var statsResetMetric float64
 		if !statsReset.Valid {
 			level.Debug(c.log).Log("msg", "Skipping collecting metric because it has no stats_reset")
-			statsResetMetric = 0.0
-		} else {
+		if statsReset.Valid {
 			statsResetMetric = float64(statsReset.Time.Unix())
 		}
 
--- a/collector/pg_stat_database.go
+++ b/collector/pg_stat_database.go
@@ -345,10 +345,12 @@ func (c *PGStatDatabaseCollector) Update(ctx context.Context, instance *instance
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


Sticksman:
cc @sysadmind @SuperQ

collector/pg_stat_database.go
```diff
@@ -344,9 +344,13 @@ func (c *PGStatDatabaseCollector) Update(ctx context.Context, instance *instance
 			level.Debug(c.log).Log("msg", "Skipping collecting metric because it has no blk_write_time")
 			continue
 		}
+
+		var statsResetMetric float64
```
SuperQ, line 348:
```suggestion
		statsResetMetric := 0.0
```

collector/pg_stat_database.go
```diff
@@ -344,9 +344,13 @@ func (c *PGStatDatabaseCollector) Update(ctx context.Context, instance *instance
...
 		if !statsReset.Valid {
 			level.Debug(c.log).Log("msg", "Skipping collecting metric because it has no stats_reset")
```
SuperQ, line 350: We're not skipping collecting anymore.

collector/pg_stat_database.go
```diff
@@ -344,9 +344,13 @@ func (c *PGStatDatabaseCollector) Update(ctx context.Context, instance *instance
...
-			continue
+			statsResetMetric = 0.0
+		} else {
+			statsResetMetric = float64(statsReset.Time.Unix())
```
SuperQ, line 353:
```suggestion
		if statsReset.Valid {
			statsResetMetric = float64(statsReset.Time.Unix())
```

collector/pg_stat_database.go
Sticksman: Yep, whoops

sysadmind (approved):

PR #878, opened 2023-07-28
Bugfix: Make sure all collectors have loggers

Sticksman:
I think this will make sure all loggers will also tell us when they got queried.

### DIFF
--- a/collector/pg_postmaster.go
+++ b/collector/pg_postmaster.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"database/sql"
 
+	"github.com/go-kit/log"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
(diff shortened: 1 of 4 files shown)


Sticksman:
@sysadmind @SuperQ

sysadmind (commented):
What code suggests that this will log when a collector gets called? I think I mist be missing something.

SuperQ:
I agree, I'm not sure I see where the `log` collector struct field is used anywhere. I wonder if this is an old feature that has been refactored away.

SuperQ:
Ahh, no, a few collectors use `c.log` for debug messages.

sysadmind:
Unless this PR is updated to include log messages when relevant, I don't think we need this PR. @Sticksman I will leave it up to you to decide if you want to add log messages or close the PR.

ArthurSens:
Hey, I'm taking the liberty to close this once since it has been a very long time without activity.

Feel free to reopen if you'd like to continue and address Joe's concerns

Issue #385, opened 2020-03-20
pg_replication_lag is invalid for postgres newer than 9.3


akamensky:
Releases after postgresql 9.3 had a change in logic where it would have no WAL updates while master host is idle, which would result in linear lag values increase when master is idle and use old query: `SELECT EXTRACT(EPOCH FROM (now() - pg_last_xact_replay_timestamp())) as lag`

Appropriate query should be checking if we are on the latest know WAL position and if not then use this query, thus correct query is something like below: `SELECT CASE WHEN pg_last_wal_receive_lsn() = pg_last_wal_replay_lsn() THEN make_interval(0,0,0,0,0,0,0.0) ELSE (now() - pg_last_xact_replay_timestamp()) END AS replication_lag;`

foozmeat:
When I tried to use the recommended query I got a parsing error on the `00:00:00` result. This query worked better for me:

`SELECT CASE WHEN pg_last_wal_receive_lsn() = pg_last_wal_replay_lsn() THEN 0 ELSE EXTRACT (epoch from (now() - pg_last_xact_replay_timestamp())) END AS lag;`

akamensky:
FWIW, not sure about earlier versions but 10/11/12 you can get replication lag from primary node without any math in it. It is directly available there in microseconds value.

geekq:
@akamensky You mean `select * from pg_stat_replication;` on the primary instance?

I have postgresql 12.5 here and implemented it as
```
pg_replication:
  query: "select extract(epoch from max(coalesce(replay_lag, make_interval(0,0,0,0,0,0,0.0)))) as lag_seconds from pg_stat_replication"
  master: true
  metrics:
    - lag_seconds:
        usage: "GAUGE"
        description: "Replication lag behind primary in seconds, as detected after a roundtrip on the primary"

```

### Explanation

From postgres documentation https://www.postgresql.org/docs/12/monitoring-stats.html#PG-STAT-REPLICATION-VIEW

> The pg_stat_replication view will contain one row per WAL sender process,
showing statistics about replication to that sender's connected standby
server. Only directly connected standbys are listed; no information is
available about downstream standby servers.

> The lag times reported in the pg_stat_replication view are measurements of
the time taken for recent WAL to be written, flushed and replayed and for the
sender to know about it. These times represent the commit delay that was (or
would have been) introduced by each synchronous commit level, if the remote
server was configured as a synchronous standby. For an asynchronous standby,
[...]

PR #895, opened 2023-08-28
fix pg_replication_lag_seconds

IamLuksha:
Since postgres_exporter is being tested for versions above 10, I suggest replacing 
```
                WHEN NOT pg_is_in_recovery() THEN 0
		ELSE GREATEST (0, EXTRACT(EPOCH FROM (now() - pg_last_xact_replay_timestamp())))
```
to
```
		WHEN NOT pg_is_in_recovery() THEN 0
                WHEN pg_last_wal_receive_lsn () = pg_last_wal_replay_lsn () THEN 0
		ELSE GREATEST (0, EXTRACT(EPOCH FROM (now() - pg_last_xact_replay_timestamp())))
```

Changed the way `pg_last_xact_replay_timestamp` works after 9.3

https://github.com/prometheus-community/postgres_exporter/issues/385

### DIFF
--- a/collector/pg_replication.go
+++ b/collector/pg_replication.go
@@ -55,6 +55,7 @@ var (
 	pgReplicationQuery = `SELECT
 	CASE
 		WHEN NOT pg_is_in_recovery() THEN 0
+                WHEN pg_last_wal_receive_lsn () = pg_last_wal_replay_lsn () THEN 0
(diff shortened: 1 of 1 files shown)


SuperQ:
:eyeglasses: @Sticksman

abh:
Merging this in made the metric much more useful for me (we have some data bases that sometimes go many minutes without a change).

sysadmind (approved):

sysadmind:
LGTM. Thanks!

Issue #898, opened 2023-08-31
Connections are leaking


MTRNord:
**What did you do?**

I am running the prometheus-community helmchart. No autodiscovery, no config, one DSN setup. Nothing unusual afaik.

**What did you expect to see?**

I expected it to go up in connections and then hit a ceiling.

**What did you see instead? Under which circumstances?**

![image](https://github.com/prometheus-community/postgres_exporter/assets/1374914/504d186a-3d3d-4a94-a5ac-61a9782f9e4c)

The connections went up to 50 until pg ran out of connections (total at pg is 240 connections + 3 superuser connections):

> ts=2023-08-31T18:39:22.792Z caller=collector.go:173 level=error msg="Error opening connection to database" err="error querying postgresql version: pq: sorry, too many clients already"

**Environment**

* System information:

	Linux 6.1.31-gentoo-x86_64 x86_64 (inside kubernetes pod using the docker image provided in this repo)

* postgres_exporter version:

	postgres_exporter, version 0.13.1 (branch: master, revision: 589087912638d738d1dfeaffbace77660dc3e52e)
	build user:       root@43776d249d3b
	build date:       20230825-09:22:12
	go version:       go1.20.7
 	platform:         linux/amd64
	tags:             netgo static_build

* postgres_exporter flags:

```
--config.file=/etc/postgres_exporter.yml --web.listen-address=:9187
```

* PostgreSQL version:

	matrix-postgres-cluster-0:/home/postgres# postgres --version
	postgres (PostgreSQL) 15.2 (Ubuntu 15.2-1.pgdg22.04+1)

* Logs:
```
ts=2023-08-31T18:39:22.792Z caller=collector.go:173 level=error msg="Error opening connection to database" err="error querying postgresql version: pq: sorry, too many clients already
```
(this is repeated a lot. But otherwise no logs after startup)

sysadmind:
This was fixed in #902

PR #882, opened 2023-08-02
Delay database connection until scrape

sysadmind:
This no longer returns an error when creating a collector.instance when the database cannot be reached for the version query. This will resolve the entire postgresCollector not being registered for metrics collection when a database is not available. If the version query fails, the scrape will fail.
    
Resolves #880

PR #902, opened 2023-09-05
Fix a connection leak

albix:
I guess the leak was introduced in PR #882 and reported in #898


sysadmind (approved):
LGTM. Tha la for catching this.

PR #903, opened 2023-09-08
Fix cross-compilation command in README.md

divergentdave:
The cross-compilation command currently lists amd64 twice. This fixes it by changing the second to arm64.


sysadmind (approved):

sysadmind:
LGTM Thanks!

PR #904, opened 2023-09-08
stat_user_tables: Add total size metric

divergentdave:
This adds a new `pg_stat_user_tables_total_size` metric to the stat_user_tables collector. It uses the [`pg_total_relation_size()`](https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADMIN-DBSIZE) function. (I had been using this via a modification to the query YAML file previously)

Note that #847 does something similar, but as a separate collector. This query is only scoped to user tables, not all relations.

### DIFF
--- a/collector/pg_stat_user_tables.go
+++ b/collector/pg_stat_user_tables.go
@@ -150,6 +150,12 @@ var (
 		[]string{"datname", "schemaname", "relname"},
 		prometheus.Labels{},
 	)
+	statUserTablesTotalSize = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, userTableSubsystem, "size_bytes"),
+		"Total disk space used by this table, in bytes, including all indexes and TOAST data",
+		[]string{"datname", "schemaname", "relname"},
+		prometheus.Labels{},
+	)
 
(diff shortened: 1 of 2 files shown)


collector/pg_stat_user_tables.go
```diff
@@ -150,6 +150,12 @@ var (
 		[]string{"datname", "schemaname", "relname"},
 		prometheus.Labels{},
 	)
+	statUserTablesTotalSize = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, userTableSubsystem, "total_size"),
```
SuperQ, line 154:
```suggestion
		prometheus.BuildFQName(namespace, userTableSubsystem, "size_bytes"),
```

SuperQ (changes requested):
Minor nit, otherwise LGTM

Sticksman (approved):

Sticksman:
LGTM

SuperQ (approved):

sysadmind (approved):
LGTM. I think in the future we need to come back and think about metrics when the column is NULL. I'm not sure this metric makes sense when the value exported is 0

Issue #908, opened 2023-09-12
Image `latest` seems to not contain additional collectors


JosefMachytkaNetApp:
Could you please check process building docker images? Looks like image `latest` does not contain additional collectors. Definitely not WAL related ones.

* When I use `latest` image, WAL metrics are NOT there and if I try "--collector.wal" parameter it says unknown parameter. Also when I try help or help-long inside container running official image, it does NOT show any lines related to WAL metrics.

* But image `master` HAS wal metrics - help-long shows it and --collector.wal works here.

Image `latest` shows version:
```
postgres_exporter, version 0.13.2 (branch: HEAD, revision: 8c3604b85e38ae7141e84ecdc318b6015a196c97)
  build user:       root@2ea2be721819
  build date:       20230724-00:08:19
  go version:       go1.20.6
  platform:         linux/amd64
  tags:             netgo static_build
```

Image `master` shows version:
```
postgres_exporter, version 0.13.1 (branch: master, revision: 31ef4ed5a2c3835bca7da0ac0a77a516af472e26)
  build user:       root@33ec95979028
  build date:       20230912-13:08:56
  go version:       go1.20.8
  platform:         linux/amd64
  tags:             netgo static_build
```

Thank you very much for your effort.

JosefMachytkaNetApp:
BTW, there is another interesting problem with `master` build - with PostgreSQL 15 exporter opens still new and new connections into database, until it reaches max_connections limit.

JosefMachytkaNetApp:
Thank you, sounds good. Looking forward to test new stable release.
Again - thank you very much for all your effort.

PR #910, opened 2023-09-13
Fix bugs mentioned in #908

sysadmind:
These collectors are disabled by default, so unless enabled, they are not tested regularly.

### DIFF
--- a/collector/pg_stat_walreceiver.go
+++ b/collector/pg_stat_walreceiver.go
@@ -127,7 +127,6 @@ func (c *PGStatWalReceiverCollector) Update(ctx context.Context, instance *insta
 		return err
 	}
 
-	defer hasFlushedLSNRows.Close()
 	hasFlushedLSN := hasFlushedLSNRows.Next()
 	var query string
 	if hasFlushedLSN {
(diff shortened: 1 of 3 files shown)


collector/pg_stat_walreceiver.go
```diff
@@ -107,13 +107,13 @@ var (
 		trim(both '''' from substring(conninfo from 'host=([^ ]*)')) as upstream_host,
 		slot_name,
 		status,
-		(receive_start_lsn- '0/0') % (2^52)::bigint as receive_start_lsn,
```
sysadmind, line 110: These are used in fmt.Sprintf below and I think the % is causing the problem. I can't actually get this collector to finish running the query against my local docker.

divergentdave: I was able to reproduce this issue, and I found that calling `hasFlushedLSNRows.Close()` before running the second query fixes it. Presumably there is some deadlock waiting for one cursor to be closed before opening another on the same connection.

SuperQ: Probably worth switching these to prepared statements.

SuperQ (approved):

Issue #917, opened 2023-09-18
Multi target and  5xx


linuxmanju:
**What did you do?**

Configured latest stable release of postgres_exporter, Created an auth module.

```
auth_modules:
  foo1: # Set this to any name you want
    type: userpass
    userpass:
      username: 'USN'
      password: 'pass@123'
    options:
      # options become key=value parameters of the DSN
      sslmode: disable
```

Promtheus scrape job

```
 - job_name: 'postgres-gcp-sd'
    static_configs:
      - targets:
        - 10.180.56.18:5432
          #gce_sd_configs:
          # - project: josh-prod-392409
          # zone: asia-south1-b
          # port: 5432
    metrics_path: /probe
    params:
        auth_module: [foo1]
    relabel_configs:
      - source_labels: [__address__]
        target_label: __param_target
      - source_labels: [__param_target]
        target_label: instance
      - target_label: __address__
        replacement: 10.180.191.203:6262
  ```      

**What did you expect to see?**

Scrape to happen successfully with metrics being available in prometheus.

**What did you see instead? Under which circumstances?**

```
server returned HTTP status 500 Internal Server Error
```

**Environment**

prod

* System information:

Linux 5.10.0-23-cloud-amd64 x86_64

* postgres_exporter version:

	postgres_exporter, version 0.13.2 (branch: HEAD, revision: 8c3604b85e38ae7141e84ecdc318b6015a196c97)
  build user:       root@2ea2be721819
  build date:       20230724-00:08:19
  go version:       go1.20.6
  platform:         linux/amd64
  tags:             netgo static_build

* postgres_exporter flags:

```
 postgres_exporter  --config.file=/tmp/pg.conf  --log.level=debug --web.listen-address=:6262
```

* PostgreSQL version:

14.0

* Logs:
```
postgres_exporter  --config.file=/tmp/pg.conf  --log.level=debug --web.listen-address=:6262
ts=2023-09-18T08:12:21.675Z caller=proc.go:250 msg="Excluded databases" databases=[]
ts=2023-09-18T08:12:21.675Z caller=main.go:142 level=warn msg="Failed to create PostgresCollector" err="empty dsn"
ts=2023-09-18T08:12:21.676Z caller=tls_config.go:274 level=info msg="Listening on" address=[::]:6262
ts=2023-09-18T08:12:21.676Z caller=tls_config.go:277 level=info msg="TLS is disabled." http2=false address=[::]:6262


```

[...]

PR #918, opened 2023-09-20
Add error log when probe collector creation fails

sysadmind:
This would help debug a scenario like #917


SuperQ (approved):
We usually prefer `Debug()` level for things that can happen on every scrape. But this is probably a hard enough error that we should log it.

PR #911, opened 2023-09-13
Rework postgres_mixin dashboard

sysadmind:
Rework the postgres_mixing dashboard to be more composable. The goal is for this to be more maintainable long term. I don't know jsonnet very well, but following other projects, this appears to be in line.

This replaces the postgres-overview.json dashboard with an overview.json dashboard with the same panels. While the dashboard does not match perfectly, it does include the same data but with the correct metrics.

PR #922, opened 2023-09-22
Fix wrong metrics pg_stat_bgwriter_buffers_* names in Grafana dashboard

romankspb:


sysadmind:
The plan is to rework this whole dashboard via #911

lyy910203 (approved):
yes buffer 
image

PR #923, opened 2023-09-25
Feature. Append query label

fireflg:
Using the query label, we can visualize the number of queries in the database. Like that:
image
Promql example:
image


collector/pg_stat_statements.go
```diff
@@ -40,41 +40,42 @@ func NewPGStatStatementsCollector(config collectorConfig) (Collector, error) {
...
-		[]string{"user", "datname", "queryid"},
+		[]string{"user", "datname", "queryid", "query"},
 		prometheus.Labels{},
 	)
 
 	pgStatStatementsQuery = `SELECT
 		pg_get_userbyid(userid) as user,
 		pg_database.datname,
 		pg_stat_statements.queryid,
+		pg_stat_statements.query as query,
```
SuperQ, line 78: I would highly recommend we limit the length of this string. The stored query could be _very_ long.

We should add a flag like this:
```
--collector.stat_statements.query_text_limit=120
```

Then we can do something like this:
```suggestion
		LEFT(pg_stat_statements.query, %d) as query,
```

SuperQ (changes requested):

SuperQ:
This needs a DCO sign-off. You can use `git commit -s --amend` to add it.

CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
+## 0.14.0 / 2023-09-11
```
SuperQ, line 1: Looks like an incorrect merge happened.

SuperQ (changes requested):
The flag still needs to be added and the query formatting done.

collector/pg_stat_statements.go
sandevgo: @fireflg asked me to help with this. I've looked at the code, this feature could lead to significant refactoring, as there is no way to set individual options to collectors. Can we make the constant limit for now?

fireflg: @SuperQ

przemek-pokrywka: @SuperQ I would also think that a hardcoded limit might meet the needs of many users already, provided that a significant refactoring would be necessary to pass the flag to the collector.

SuperQ (changes requested):
The commit history and conflicts need to be fixed.

fireflg:
> The commit history and conflicts need to be fixed.

Check please

przemek-pokrywka:
> https://github.com/prometheus-community/postgres_exporter/pull/923#issuecomment-1880106853

@SuperQ  please :pray:

quolpr:
@SuperQ could you check the PR please? It's super important, hard to debug with just queryid

ArthurSens:
Hey, sorry for the delayed review here. It looks like #940 already solved the same problem, so I'll close this one.

Please do reopen if you think I misunderstood the intention of this PR!

PR #939, opened 2023-10-24
Allow enable/disable collectors via environment variables.

Delorien84:


SuperQ (changes requested):
Sorry, we don't typically use environment variables for configuration. This is what flags are for.

Delorien84:
Yes but there are situation, when command line arguments cannot be set for example https://github.com/zalando/postgres-operator/issues/2439

PR #919, opened 2023-09-21
Fix test build failures on 32-bit arch again

dswarbrick:
Another case of untyped integer overflows on 32-bit arch.

Fixes failed test build:
```
$ GOARCH=386 go test ./...
# github.com/prometheus-community/postgres_exporter/collector [github.com/prometheus-community/postgres_exporter/collector.test]
collector/pg_stat_walreceiver_test.go:67:4: cannot use 1200668684563608 (untyped int constant) as int value in argument to sqlmock.NewRows(columns).AddRow (overflows)
collector/pg_stat_walreceiver_test.go:69:4: cannot use 1200668684563609 (untyped int constant) as int value in argument to sqlmock.NewRows(columns).AddRow (overflows)
collector/pg_stat_walreceiver_test.go:73:4: cannot use 1200668684563610 (untyped int constant) as int value in argument to sqlmock.NewRows(columns).AddRow (overflows)
collector/pg_stat_walreceiver_test.go:146:4: cannot use 1200668684563608 (untyped int constant) as int value in argument to sqlmock.NewRows(columns).AddRow (overflows)
collector/pg_stat_walreceiver_test.go:151:4: cannot use 1200668684563610 (untyped int constant) as int value in argument to sqlmock.NewRows(columns).AddRow (overflows)
ok      github.com/prometheus-community/postgres_exporter/cmd/postgres_exporter 0.007s
FAIL    github.com/prometheus-community/postgres_exporter/collector [build failed]
ok      github.com/prometheus-community/postgres_exporter/config        0.005s
FAIL
```

Please consider adding 32-bit builds / tests to CI, so that these are caught automatically in future.

Refs: #844

PR #944, opened 2023-10-27
Release v0.15.0

SuperQ:
* [ENHANCEMENT] Add 1kB and 2kB units #915
* [BUGFIX] Add error log when probe collector creation fails #918
* [BUGFIX] Fix test build failures on 32-bit arch #919
* [BUGFIX] Adjust collector to use separate connection per scrape #936


SuperQ:
PG 10 was EoL 1 year ago.

https://www.postgresql.org/support/versioning/

sysadmind (approved):
LGTM. Is the policy that we will not intentionally break compat with V10 but without any guarantees?

SuperQ:
Yea, that's the policy we've been sticking to. It's just good to remove old versions from the pipeline to not waste resources.

roock:
@SuperQ Container images have been published on docker hub and quay.io, but I don't see the new version on the Github Release page. Will it be pushed there later or is/was there an error in the release process?

SuperQ:
@roock Thanks for the reminder, I forgot to mark the draft release as published. Done.

PR #954, opened 2023-11-11
feat: add usename to pg_process_idle metric

rezaxd:
Add usename to query of pg_process_idle metric.

### DIFF
--- a/collector/pg_process_idle.go
+++ b/collector/pg_process_idle.go
@@ -52,11 +52,12 @@ func (PGProcessIdleCollector) Update(ctx context.Context, instance *instance, ch
 				SELECT
 				state,
 				application_name,
+				usename,
 				SUM(EXTRACT(EPOCH FROM (CURRENT_TIMESTAMP - state_change))::bigint)::float AS process_idle_seconds_sum,
 				COUNT(*) AS process_idle_seconds_count
 				FROM pg_stat_activity
 				WHERE state ~ '^idle'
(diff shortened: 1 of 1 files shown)


collector/pg_process_idle.go
```diff
@@ -52,11 +52,12 @@ func (PGProcessIdleCollector) Update(ctx context.Context, instance *instance, ch
 				SELECT
 				state,
 				application_name,
+    			usename,
```
sysadmind, line 55: The indentation is not correct here.

sysadmind (changes requested):
The database scanning wasn't updated so this will fail and there is no label changes. Is the intention to add a label to the metric?

collector/pg_process_idle.go
rezaxd: Sorry for that. But I fixed that in force-push. Am I right?
Thank for your time.

rezaxd:
`usename` added to labels now.

rezaxd:
Hello dear @sysadmind !
Fixed my mistakes.

ArthurSens:
Hey, sorry for the very late review. I'll be closing the PR because the codebase has changed a lot since the last activity here, and important things are missing, such as test coverage; more importantly, we would like to understand the motivation for adding this new label. Is this enabling any new use case that isn't possible today? Could we discuss that in an Issue before jumping into implementation?

PR #961, opened 2023-11-16
pg_stat_database: added support for `active_time` counter

sitole:
Hello,
for our use case is key to measure active time per database and this metric was missing in current version of `pg_exporter`.
In older version we add this metric manually via `queries.yaml` but this method is now not recommended.

### DIFF
--- a/collector/pg_stat_database.go
+++ b/collector/pg_stat_database.go
@@ -209,9 +209,9 @@ var (
 	statDatabaseActiveTime = prometheus.NewDesc(prometheus.BuildFQName(
 		namespace,
 		statDatabaseSubsystem,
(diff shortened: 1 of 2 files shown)


sitole:
hey guys, should be fully tested. Can you check it? @SuperQ @sysadmind

SuperQ (approved):
Nice, thanks!

Never-NG:
`active_time` column doesn't exist before PG-14.0， so this change doesn't support for  postgreSQL before 14.0.
https://www.postgresql.org/docs/13/monitoring-stats.html#MONITORING-PG-STAT-DATABASE-VIEW

PR #968, opened 2023-11-28
Fixed postgresql>=10 secondary server lag always 0

Efiartes:
Fixed postgresql>=10 secondary server lag always 0, now it detects version and executes one query or the other

### DIFF
--- a/collector/pg_replication.go
+++ b/collector/pg_replication.go
@@ -15,8 +15,9 @@ package collector
 
 import (
 	"context"
-
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/blang/semver/v4"
+	"database/sql"
 )
 
 const replicationSubsystem = "replication"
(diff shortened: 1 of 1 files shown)


collector/pg_replication.go
```diff
@@ -62,13 +63,31 @@ var (
...
+	var row *sql.Row
+        if after10 >= 0 {
+		row = db.QueryRowContext(ctx,
+                        pgReplicationQueryAfterVersion10,
+                )
+        } else {
+		row = db.QueryRowContext(ctx,
+                        pgReplicationQueryBeforeVersion10,
+                )
+        }
```
SuperQ, line 90: This can be simplified a bit.
```suggestion
	query := pgReplicationQueryBeforeVersion10
	if instance.version.GE(semver.MustParse("10.0.0")) {
		query = pgReplicationQueryAfterVersion10
	}
	row := db.QueryRowContext(ctx, query)
```

SuperQ (commented):
A suggested improvement to the version logic.

SuperQ (changes requested):
This needs a DCO sign-off. You can use `git commit -s --amend` to add it.

PR #978, opened 2023-11-30
fix: longRunningTransactionsAgeInSeconds

roock:
- clock_timestamp will just return the current timestamp
- what we actually want to know is the lowest timestamp of the xact_start column

Please also see:
https://github.com/prometheus-community/postgres_exporter/pull/836#discussion_r1295748514
and the original can be found here:
https://gitlab.com/gitlab-cookbooks/gitlab-exporters/-/blob/master/templates/postgres_exporter/queries.yaml.erb#L954

(the originally proposed version `MAX(EXTRACT(EPOCH FROM (clock_timestamp() - xact_start))) AS age_in_seconds` is also correct, but return the seconds, not the timestamp.

Using the current version the graphs look like this (flat line as this will always match the current time):
Screenshot 2023-11-30 at 15 51 56

### DIFF
--- a/collector/pg_long_running_transactions.go
+++ b/collector/pg_long_running_transactions.go
@@ -52,7 +52,7 @@ var (
 	longRunningTransactionsQuery = `
 	SELECT
 		COUNT(*) as transactions,
-   		MAX(EXTRACT(EPOCH FROM clock_timestamp())) AS oldest_timestamp_seconds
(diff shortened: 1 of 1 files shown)


sberlin:
The collector for long running transactions is not telling anything about long running transactions. Can we get this reviewed and merged, please?

roock:
Closing in favor of #1092

Issue #985, opened 2023-12-11
add client_addr to pg_stat_activity_max_tx_duration metric


betanummeric:
## proposal
Add the `client_addr` field from `pg_stat_activity` as a label to the metric `pg_stat_activity_max_tx_duration`.

## use case
My company runs the postgres_exporter for production PostgreSQL instances. Often the application (postgres client) is running on multiple hosts, sharing a common username. When one application instance blocks others by keeping an idle transaction open for too long, we need the `client_addr` information to figure out which application host is causing that.
By adding the `client_addr` field to the metrics, we would enable the application admins to determine the malfunctioning host without needing to ask a DBA (someone with privileges to query `pg_stat_activity` directly).

PR #986, opened 2023-12-11
add client_addr to metric pg_stat_activity_max_tx_duration

betanummeric:
implements #985

### DIFF
--- a/cmd/postgres_exporter/postgres_exporter.go
+++ b/cmd/postgres_exporter/postgres_exporter.go
@@ -252,6 +252,7 @@ var builtinMetricMaps = map[string]intermediateMetricMap{
 			"state":            {LABEL, "connection state", nil, semver.MustParseRange(">=9.2.0")},
 			"usename":          {LABEL, "connection usename", nil, nil},
 			"application_name": {LABEL, "connection application_name", nil, nil},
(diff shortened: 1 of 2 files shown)


SuperQ (commented):
Hmm, this could impact cardinality very badly as things scale up. I'm not sure this would be acceptable for most users.

betanummeric:
@SuperQ I don't think most users have so much more client hosts than postgres-users to make the `client_addr`-label problematic. But sure, this introduces another factor to the amount of monitoring data.
Would it be ok for you to make this change opt-in using a CLI flag?

SuperQ:
An opt-in flag would be fine. The only issue is that this collector setup hasn't been moved to our new collector package, so it's possibly slightly difficult to implement here.

sysadmind:
I don't know that we have had a good example of customizing a collector that has been ported to the collector package either. I remember another request noting that the constructors for collectors don't have an obvious pattern for customization. @SuperQ Do we have a good example from another exporter for customizing collectors?

ArthurSens:
Hey, I'm taking the liberty of closing this PR since it's been a while without activity.

The codebase this PR touched has moved to a separate collector implementation, and we have examples of collectors using extra configuration options that could be good references for anyone who wants to continue this PR!

PR #1003, opened 2024-02-12
use Info level for excluded databases log message

keegancsmith:
This is the only log message which didn't specify a level in the postgres_exporter. I am unsure if this log message should be info or debug, but leaning towards the more important since previously it would just always log.

The way I validated this was the only non-leveled logger was via grep. Both of these only returned this callsite previously:

``` shell
git grep 'logger\.Log'
git grep '\.Log(' | grep -v level
```

### DIFF
--- a/cmd/postgres_exporter/main.go
+++ b/cmd/postgres_exporter/main.go
@@ -93,7 +93,7 @@ func main() {
 	}
 
 	excludedDatabases := strings.Split(*excludeDatabases, ",")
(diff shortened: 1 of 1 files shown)


keegancsmith:
I saw that the last PR that wasn't made by a bot did a ping on some people to get it reviewed. So hope you don't mind me pinging ya for this tiny change @SuperQ @sysadmind

sysadmind (approved):
LGTM

PR #673, opened 2022-08-15
Capture usename and application_name for pg_stat_activity

cezmunsta:
It is necessary to be able to exclude backups from long-running
transaction alerts, as they are to be expected. With the current
pg_stat_activity metric there is no ability to filter out
specific users or application names.

Resolves #668

PR #1008, opened 2024-02-26
Add usename and application_name to pg_locks_count

killzoner:
Hi,

We have a use case where several application are using the same database for update, and currently there is no easy way to determine which process is causing most of the locks.

This PR adds `usename` and `application_name` to `pg_locks_count`, in the same spirit as https://github.com/prometheus-community/postgres_exporter/pull/673

### DIFF
--- a/collector/pg_locks.go
+++ b/collector/pg_locks.go
@@ -45,13 +45,15 @@ var (
 			"count",
 		),
 		"Number of locks",
-		[]string{"datname", "mode"}, nil,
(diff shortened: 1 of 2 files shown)


killzoner:
@SuperQ sorry to jump here, just wanted to know if there is interest or if i'd rather close this :pray:

killzoner:
@sysadmind maybe?

killzoner:
Closing as no answer for months now

Issue #1017, opened 2024-04-03
Export safe_wal_size and wal_status to monitor for lost replication slots when using max_slot_wal_keep_size


MarcWort:
## Proposal
We set the option [max_slot_wal_keep_size](https://www.postgresql.org/docs/13/runtime-config-replication.html#GUC-MAX-SLOT-WAL-KEEP-SIZE) supported by postgres >= 13 and want to alert if the slot is lost. I cannot find a reliable solution to do this except using `wal_status` and `safe_wal_size` which is currently not exported.

documentation: https://www.postgresql.org/docs/13/view-pg-replication-slots.html

example:
```
postgres=# SELECT * FROM pg_catalog.pg_replication_slots;
 slot_name | plugin | slot_type | datoid | database | temporary | active | active_pid | xmin | catalog_xmin | restart_lsn | confirmed_flush_lsn | wal_status | safe_wal_size
-----------+--------+-----------+--------+----------+-----------+--------+------------+------+--------------+-------------+---------------------+------------+---------------
 barman    |        | physical  |        |          | f         | f      |            |      |              | 0/B0000000  |                     | unreserved |    -323906992
(1 rows)

postgres=# SELECT * FROM pg_catalog.pg_replication_slots;
 slot_name | plugin | slot_type | datoid | database | temporary | active | active_pid | xmin | catalog_xmin | restart_lsn | confirmed_flush_lsn | wal_status | safe_wal_size
-----------+--------+-----------+--------+----------+-----------+--------+------------+------+--------------+-------------+---------------------+------------+---------------
 barman    |        | physical  |        |          | f         | f      |            |      |              |             |                     | lost       |
(1 rows)
```

sysadmind:
Added in #1027 (v0.16.0): `pg_replication_slots_safe_wal_size_bytes` and `..._wal_status`. Closing as resolved.

PR #1027, opened 2024-05-07
feat: Add safe_wal_size and wal_status to replication_slot 

MarcWort:
implements #1017

### DIFF
--- a/collector/pg_replication_slot.go
+++ b/collector/pg_replication_slot.go
@@ -72,6 +72,15 @@ var (
 		"number of bytes that can be written to WAL such that this slot is not in danger of getting in state lost",
 		[]string{"slot_name", "slot_type"}, nil,
 	)
+	pgReplicationSlotWalStatus = prometheus.NewDesc(
+		prometheus.BuildFQName(
+			namespace,
+			replicationSlotSubsystem,
+			"wal_status",
+		),
+		"availability of WAL files claimed by this slot",
+		[]string{"slot_name", "slot_type", "wal_status"}, nil,
+	)
 
 	pgReplicationSlotQuery = `SELECT
 		slot_name,
@@ -83,7 +92,8 @@ var (
 		END AS current_wal_lsn,
 		COALESCE(confirmed_flush_lsn, '0/0') - '0/0' AS confirmed_flush_lsn,
 		active,
-		safe_wal_size
+		safe_wal_size,
+		wal_status
 	FROM pg_replication_slots;`
 )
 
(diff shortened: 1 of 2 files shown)


collector/pg_replication_slot.go
```diff
@@ -63,6 +63,24 @@ var (
 		"whether the replication slot is active or not",
 		[]string{"slot_name", "slot_type"}, nil,
 	)
+	pgReplicationSlotSafeWal = prometheus.NewDesc(
+		prometheus.BuildFQName(
+			namespace,
+			replicationSlotSubsystem,
+			"safe_wal_size",
```
SuperQ, line 70:
```suggestion
			"safe_wal_size_bytes",
```

collector/pg_replication_slot.go
```diff
@@ -131,6 +153,28 @@ func (PGReplicationSlotCollector) Update(ctx context.Context, instance *instance
...
+			)
+		}
+
+		if walStatus.Valid {
+			// See https://www.postgresql.org/docs/14/view-pg-replication-slots.html
+			walStatusMap := map[string]int{
+				"reserved":   0, // reserved means that the claimed files are within max_wal_size.
+				"extended":   1, // extended means that max_wal_size is exceeded but the files are still retained, either by the replication slot or by wal_keep_size.
+				"unreserved": 2, // unreserved means that the slot no longer retains the required WAL files and some of them are to be removed at the next checkpoint. This state can return to reserved or extended.
+				"lost":       3, // lost means that some required WAL files have been removed and this slot is no longer usable.
```
SuperQ, line 170: We generally try to avoid enum style metrics in Prometheus. We don't really have a first-class enum concept. (I wish we did)

There are two options I would recommend:
* Have a `wal_status` metric with a `wal_status="FOO"` metric with a value of 1.
* Have a `wal_status` metric with all valid statuses, but boolean for each state.
Or,

SuperQ (changes requested):
Thanks, good so far, but needs some changes.

collector/pg_replication_slot.go
MarcWort: yeah really sad. i was hoping to do a wal_status > 0 alert, but have to agree its hard to document theses numbers meaning. Anyway, I followed your first suggestion.

SuperQ: With the state set mode, you can do an alert like `sum without (wal_status) (pg_replication_wal_status) > 0`. So that should work exactly as you want.

SuperQ: With the current method, your alert would be also simple:

```
pg_replication_wal_status{wal_status!="reserved"} == 1
```

This has the advantage of also providing the status as a string to your alert labels template.

SuperQ (approved):
LGTM

mbanck-ntap:
`safe_wal_size` was added to `pg_replication_slots` in version 13, how is this handled on version 12 servers?

mbanck-ntap:
Also, why are the `pg_replication_slots_*` (not `pg_replication_slot_*`) metrics from `cmd/postgres_exporter/postgres_exporter.go` still around, are they not a duplicate of `collector/pg_replication_slot.go`?

collector/pg_replication_slot.go
```diff
@@ -73,7 +91,9 @@ var (
 		    pg_current_wal_lsn() - '0/0'
 		END AS current_wal_lsn,
 		COALESCE(confirmed_flush_lsn, '0/0') - '0/0' AS confirmed_flush_lsn,
-		active
+		active,
+		safe_wal_size,
+		wal_status
```
ringerc, line 96: These were added in postgres 13, they are absent in 12 and below, so this will break scraping on Pg 12. See https://www.postgresql.org/docs/12/view-pg-replication-slots.html

Issue #1044, opened 2024-06-11
Column active_time missing in postgres < 14


MarcWort:
As @dbakit already said in https://github.com/prometheus-community/postgres_exporter/pull/961#issuecomment-1855052713 `active_time` is only available on postgres >=14. 

Maybe there is a need for versioning pg versions in the collector package?

sysadmind:
There is support for versions in the collector package. The instance struct has the postgres version for comparison. 

https://github.com/prometheus-community/postgres_exporter/blob/master/collector/instance.go

PR #1045, opened 2024-06-14
fix: Only query active_time on pg>=14

MarcWort:
Fixes https://github.com/prometheus-community/postgres_exporter/issues/1044

Its quite hacky IMO, but please let me know what you think or how we could solve this otherwise.


sysadmind (approved):
LGTM. I would love to see an additional test for a version < 14, but I think this is sound.

Issue #1059, opened 2024-08-12
pg_stat


RavinaChidambaram:
## Proposal
**Use case. Why is this important?**

*“Nice to have” is not a good use case. :)*

PR #1067, opened 2024-09-09
Added support for fetching pg_stat_user_tables metrics from all databases 

RavinaChidambaram:
Implemented #1059
Metrics from pg_stat_user_tables were previously collected only from the postgres database. Given the various use cases and the significance of metrics from this table across all available databases, I have updated the code to retrieve metrics from pg_stat_user_tables for all databases.

### DIFF
--- a/collector/pg_stat_user_tables.go
+++ b/collector/pg_stat_user_tables.go
@@ -16,7 +16,10 @@ package collector
 import (
 	"context"
 	"database/sql"
-
+	"fmt"
+	"os"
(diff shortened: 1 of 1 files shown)


Sticksman:
Looks like there are some tests that need fixing here

ArthurSens:
Hey 👋 -- I'm taking the liberty of closing this PR due to inactivity. 

Also, a different design is being proposed in #1326, and a current implementation attempt in #1378.

PR #1070, opened 2024-09-23
fix converting NULL to uint64 is unsupported in size table

Electronn:
This patch fixes bug of unwanted null in uint64 type. Bug causes errors on AWS Aurora PostgreSQL clusters:
```
Scan error on column index 1, name \"size\": converting NULL to uint64 is unsupported
```


Sticksman (approved):

collector/pg_wal.go
```diff
@@ -66,19 +66,34 @@ func (c PGWALCollector) Update(ctx context.Context, instance *instance, ch chan<
...
-	var segments uint64
-	var size uint64
+	var segments sql.NullInt64
+	var size sql.NullInt64
 	err := row.Scan(&segments, &size)
 	if err != nil {
 		return err
 	}
+
+	var segmentsValue float64
```
sysadmind, line 76: segments should never be NULL. It's a `count(*)` which I don't believe can return a NULL value. The error you reported was on the size column. If Aurora is returning NULL for that column, I think there should not be a metric emitted from the exporter. I don't know what value a metric for WAL size with a value of 0 is.

sysadmind (changes requested):

ArthurSens:
Hey, I'm closing this PR because there's been no activity for a long time. It would be awesome to have this fixed as part of the work @megative is doing to better support Aurora!

PR #1073, opened 2024-10-15
chore!: adopt log/slog, drop go-kit/log

tjhop:
Changes included:
- update go to 1.23
- bump deps for prometheus/{client_golang,common,exporter-toolkit} libs
- converts the exporter to use log/slog for logging, drops deps for go-kit/log and go-logfmt/logfmt


tjhop:
cc: @SuperQ

SuperQ (approved):

PR #1074, opened 2024-10-16
Cleanup unused code from NewPostgresCollector

ajmaidak:
Clean up a dead codepath in the NewPostgresCollector constructor.

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -97,34 +97,15 @@ type PostgresCollector struct {
 type Option func(*PostgresCollector) error
 
 // NewPostgresCollector creates a new PostgresCollector.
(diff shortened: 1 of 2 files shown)


collector/collector.go
```diff
@@ -97,34 +97,15 @@ type PostgresCollector struct {
 type Option func(*PostgresCollector) error
```
sysadmind, line 97: I think this makes the Option type no longer needed. 
```suggestion
```

ArthurSens:
Hey, I'm taking the liberty of closing this PR due to the inactivity. The comment above went unaddressed, and the codebase has changed significantly since then.

PR #1075, opened 2024-10-17
Collect idle_in_transaction_time from PG >= 15

ajmaidak:
idle_in_transaction_time was added to pg_stat_database in pg 15

https://www.postgresql.org/docs/15/monitoring-stats.html#MONITORING-PG-STAT-DATABASE-VIEW

### DIFF
--- a/collector/pg_stat_database.go
+++ b/collector/pg_stat_database.go
@@ -218,6 +218,15 @@ var (
 		[]string{"datid", "datname"},
 		prometheus.Labels{},
 	)
+	statDatabaseIdleInTransaction = prometheus.NewDesc(prometheus.BuildFQName(
+		namespace,
+		statDatabaseSubsystem,
+		"idle_in_transaction_time_seconds_total",
+	),
+		"Time spent idling while in a transaction in this database, in seconds",
+		[]string{"datid", "datname"},
+		prometheus.Labels{},
+	)
 )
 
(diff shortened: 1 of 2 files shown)


Sticksman (approved):

collector/pg_stat_database.go
```diff
@@ -375,6 +393,11 @@ func (c *PGStatDatabaseCollector) Update(ctx context.Context, instance *instance
 			continue
 		}
 
+		if idleInTransactionTimeAvail && !idleInTransactionTime.Valid {
+			level.Debug(c.log).Log("msg", "Skipping collecting metric because it has no idle_in_transaction_time")
+			continue
```
sysadmind, line 398: I don't think this needs to skip the whole row in this case. I think this logic can be to skip sending the metric to the channel down on line 540 instead. I know the logic above is this way, but I think technically this addition could break things for other users if this field is NULL, so I want to avoid that.

sysadmind (changes requested):

kyrias:
Any chance this PR will go anywhere at this point, or should I open a new one?

ArthurSens:
@kyrias feel free to open a new one. We've been waiting for the author for quite a while now

ArthurSens:
Taking the liberty to close this PR due to the amount of time without activity

PR #1077, opened 2024-10-18
Upgrade postgres exporter prometheus packages

Sticksman:
Upgrade common to latest version and use the proper version.NewCollector from golang_client


Sticksman:
👓 @SuperQ

SuperQ:
FYI, this is already being done in https://github.com/prometheus-community/postgres_exporter/pull/1073

PR #1082, opened 2024-11-02
Add pg_stat_wal table metrics

### LINKED ISSUE #1028: Add support for wal_fpi in pg_stat_wal
## Add support for `pg_stat_wal`
**Use case. Why is this important?**
I have a use case from my experience. I had huge spikes on checkpoints. At first, I thought the checkpointer was late and that I needed to increase max_wal_size/checkpoint_timeout, but it didn't help. I accidentally noticed that wal_fpi increased along with the spikes. To prove this, my workaround was to sample it with an SQL exporter, and it turned out to be true. The core problem was that we had a table with a GUID index and a huge insert workload, which frequently dirties almost all the index's pages, triggering massive wal_fpi on checkpoints. Having at least wal_fpi would be very useful to quickly discover such problems. Also, I think it would be useful to have the entire pg_stat_wal in the exporter, but for other columns, I don't have any motivation.

pandar00:
* Adds the following metrics. [pg_stat_wal table](https://www.postgresql.org/docs/16/monitoring-stats.html#MONITORING-PG-STAT-WAL-VIEW). It is off by default
```
"wal_records",
"wal_fpi",          // bigint
"wal_bytes",        // numeric
"wal_buffers_full", // bigint
"wal_write",        // bigint
"wal_sync",         // bigint
"wal_write_time",   // double precision
"wal_sync_time",    // double precision
"stats_reset",      // timestamp with time zone
```
* Mostly copied from `pg_stat_bgwriter` metric implementation because it is essentially the same except for the column names. 
* Resolves https://github.com/prometheus-community/postgres_exporter/issues/1028

```
# HELP pg_stat_wal_stats_reset Time at which these statistics were last reset
# TYPE pg_stat_wal_stats_reset counter
pg_stat_wal_stats_reset 1.730575462e+09
# HELP pg_stat_wal_wal_buffers_full Number of times WAL data was written to disk because WAL buffers became full
# TYPE pg_stat_wal_wal_buffers_full counter
pg_stat_wal_wal_buffers_full 1.040989e+06
# HELP pg_stat_wal_wal_bytes Total amount of WAL generated in bytes
# TYPE pg_stat_wal_wal_bytes counter
pg_stat_wal_wal_bytes 8.8115219412e+10
# HELP pg_stat_wal_wal_fpi Total number of WAL full page images generated
# TYPE pg_stat_wal_wal_fpi counter
pg_stat_wal_wal_fpi 1.2850296e+07
# HELP pg_stat_wal_wal_records Total number of WAL records generated
# TYPE pg_stat_wal_wal_records counter
pg_stat_wal_wal_records 3.3412574e+07
# HELP pg_stat_wal_wal_sync Number of times WAL files were synced to disk via issue_xlog_fsync request (if fsync is on and wal_sync_method is either fdatasync, fsync or fsync_writethrough, otherwise zero). See Section 30.5 for more information about the internal WAL function issue_xlog_fsync.
# TYPE pg_stat_wal_wal_sync counter
pg_stat_wal_wal_sync 74409
# HELP pg_stat_wal_wal_sync_time Total amount of time spent syncing WAL files to disk via issue_xlog_fsync request,	in milliseconds (if track_wal_io_timing is enabled, fsync is on, and wal_sync_method is either fdatasync, fsync or fsync_writethrough, otherwise zero).
# TYPE pg_stat_wal_wal_sync_time counter
pg_stat_wal_wal_sync_time 0
# HELP pg_stat_wal_wal_write Number of times WAL buffers were written out to disk via XLogWrite request. See Section 30.5 for more information about the internal WAL function XLogWrite.
# TYPE pg_stat_wal_wal_write counter
pg_stat_wal_wal_write 1.128863e+06
# HELP pg_stat_wal_wal_write_time Total amount of time spent writing WAL buffers to disk via XLogWrite request, in milliseconds (if track_wal_io_timing is enabled, otherwise zero). This includes the sync time when wal_sync_method is either open_datasync or open_sync.
# TYPE pg_stat_wal_wal_write_time counter
pg_stat_wal_wal_write_time 0
```


Sticksman (approved):

collector/pg_stat_wal.go
```diff
@@ -0,0 +1,240 @@
+// Copyright 2023 The Prometheus Authors
```
SuperQ, line 1:
```suggestion
// Copyright 2024 The Prometheus Authors
```

collector/pg_stat_wal.go
```diff
@@ -0,0 +1,240 @@
...
+type PGStatWALCollector struct {
+	log *slog.Logger
+}
+
+func NewPGStatWALCollector(config collectorConfig) (Collector, error) {
+	return &PGStatWALCollector{log: config.logger}, nil
+}
+
+var statsWALRecordsDesc = prometheus.NewDesc(
+	prometheus.BuildFQName(namespace, statWALSubsystem, "wal_records"),
```
SuperQ, line 40: Counter values must end in `_total`

```suggestion
	prometheus.BuildFQName(namespace, statWALSubsystem, "wal_records_tota"),
```

SuperQ (changes requested):

sysadmind:
Bump @SuperQ

jonathanhartley:
When can this be merged its really useful ?

collector/pg_stat_wal.go
```diff
@@ -0,0 +1,240 @@
...
+			&statsReset,
+		)
+		if err != nil {
+			return err
+		}
+
+		walRecordsMetric := 0.0
+		if walRecords.Valid {
+			walRecordsMetric = float64(walRecords.Int64)
+		}
```
ArthurSens, line 152: If the query returns NULL, we want to avoid exposing the metric instead of exposing with 0 as value, same applies to all other metrics.

collector/pg_stat_wal.go
```diff
@@ -0,0 +1,240 @@
...
+		if statsReset.Valid {
+			resetMetric = float64(statsReset.Time.Unix())
+		}
+		ch <- prometheus.MustNewConstMetric(
+			statsWALStatsResetDesc,
+			prometheus.CounterValue,
+			resetMetric,
+		)
+	}
+	return nil
```
ArthurSens, line 239: should we return rows.Err() here?

ArthurSens (changes requested):
Hi, sorry for the super-delayed review! The codebase has been heavily refactored in the last months, so we'll need a big rebase or start a fresh PR from scratch.

Also, there are a few things missing here:
* `pg_stat_wal` doesn't exist for Postgres 13 or below
* Postgres 18 moves some of these metrics to `pg_stat_io`
* We probably want integration tests to ensure we run the right queries in the right versions

We also want to follow [Prometheus naming conventions](https://prometheus.io/docs/practices/naming/), counters get a `_total` suffix, we convert milliseconds to seconds, reset_timestamp becomes a gauge like `stats_reset_timestamp_seconds`. If we can it would be awesome if we could avoid repeating `wal` in the subsystem and them again on the rest of the string that builds the metric name, to avoid a metric like `pg_stat_wal_wal_records_total`

I've also noticed that #1167 and #1169 seem almost identical to this PR, so we might want to coordinate the work :)

Issue #1007, opened 2024-02-26
Add support for pg_stat_replication.(write_lag|flush_lag|replay_lag)


spootze:
## Add support for `pg_stat_replication.(write_lag|flush_lag|replay_lag)`

**Use case. Why is this important?**

For simple primary:replica setups it would be convenient to be able to monitor standby lag in seconds from the primary. `pg_stat_replication` already includes this information on postgres>=10, but the exporter does not parse it. AFAIK, workarounds would include

- monitoring byte offset from `pg_stat_replication.write_lsn|flush_lsn|replay_lsn)`, but this does not quite capture issues where the replica has not replayed a business critical change however small in bytes.
- monitoring delay on the replica, but this would require spinning up a separate monitor only to monitor a single value on the replica.

**Notes**

I'm assuming the exporter does not support parsing the `interval` data type, which is why these metrics are marked as `DISCARD` [here](https://github.com/prometheus-community/postgres_exporter/blob/838f09c97f35aaabace3d04c170bd449bda0099e/cmd/postgres_exporter/postgres_exporter.go#L218). I wonder if the following, explicit approach would be appropriate in this case:

```diff
diff --git a/cmd/postgres_exporter/queries.go b/cmd/postgres_exporter/queries.go
index fa0b5c2..e28d7b4 100644
--- a/cmd/postgres_exporter/queries.go
+++ b/cmd/postgres_exporter/queries.go
@@ -53,7 +53,10 @@ var queryOverrides = map[string][]OverrideQuery{
 			SELECT *,
 				(case pg_is_in_recovery() when 't' then null else pg_current_wal_lsn() end) AS pg_current_wal_lsn,
 				(case pg_is_in_recovery() when 't' then null else pg_wal_lsn_diff(pg_current_wal_lsn(), pg_lsn('0/0'))::float end) AS pg_current_wal_lsn_bytes,
-				(case pg_is_in_recovery() when 't' then null else pg_wal_lsn_diff(pg_current_wal_lsn(), replay_lsn)::float end) AS pg_wal_lsn_diff
+				(case pg_is_in_recovery() when 't' then null else pg_wal_lsn_diff(pg_current_wal_lsn(), replay_lsn)::float end) AS pg_wal_lsn_diff,
+				(case pg_is_in_recovery() when 't' then null else  extract(seconds from write_lag) end) as write_lag_seconds,
[...]

PR #1085, opened 2024-11-06
Export last replay age in replication collector

bitfehler:
The exported replication lag does not handle all failure modes, and can report 0 for replicas that are out of sync and incapable of recovery.

A proper replacement for that metric would require a different approach (see e.g. #1007), but for a lot of folks, simply exporting the age of the last replay can provide a pretty strong signal for something being amiss.

I think this solution might be preferable to #977, though the lag metric needs to be fixed or abandoned eventually.

### DIFF
--- a/collector/pg_replication_test.go
+++ b/collector/pg_replication_test.go
@@ -31,9 +31,9 @@ func TestPgReplicationCollector(t *testing.T) {
 
 	inst := &instance{db: db}
 
-	columns := []string{"lag", "is_replica"}
+	columns := []string{"lag", "is_replica", "last_replay"}
 	rows := sqlmock.NewRows(columns).
-		AddRow(1000, 1)
+		AddRow(1000, 1, 3)
 	mock.ExpectQuery(sanitizeQuery(pgReplicationQuery)).WillReturnRows(rows)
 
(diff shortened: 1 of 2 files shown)


collector/pg_replication_test.go
```diff
@@ -31,9 +31,9 @@ func TestPgReplicationCollector(t *testing.T) {
 
 	inst := &instance{db: db}
 
-	columns := []string{"lag", "is_replica"}
+	columns := []string{"lag", "is_replica", "last_replay"}
 	rows := sqlmock.NewRows(columns).
-		AddRow(1000, 1)
+		AddRow(1000, 1, 3)
```
Sticksman, line 36: You should add an additional expected value below

Sticksman (changes requested):

collector/pg_replication_test.go
bitfehler: Fixed, thanks!

Sticksman (approved):

garry-t:
this can fix issue for postgres 17. since replication lag now is not reports

sysadmind (approved):

PR #1086, opened 2024-11-06
Make walreceiver collector useful w/o repmgr

bitfehler:
In a streaming replication setup that was created without replication manager (`repmgr`), the `stat_wal_receiver` collector does not return any metrics, because one value it wants to export is not present.

This is rather overly opinionated. The missing metric is comparatively uninteresting and does not justify discarding all the others. And replication setups created without `repmgr` are not exactly rare.

This commit makes the one relevant metric optional and simply skips it if the respective value cannot be determined.


SuperQ (approved):
Thanks!

PR #1089, opened 2024-11-10
Fix version header in changelog

sysadmind:
Sorry, I missed this @SuperQ. After this merges, I'll push the tag.


SuperQ (approved):
:shipit:

PR #1092, opened 2024-11-14
Update pg_long_running_transactions.go

jyothikirant-sayukth:
To extract time in seconds for pg_long_running_transactions_oldest_timestamp_seconds query which currently returns epoch time.

### DIFF
--- a/collector/pg_long_running_transactions.go
+++ b/collector/pg_long_running_transactions.go
@@ -50,11 +50,13 @@ var (
 	)
 
 	longRunningTransactionsQuery = `
-	SELECT
-		COUNT(*) as transactions,
-   		MAX(EXTRACT(EPOCH FROM clock_timestamp())) AS oldest_timestamp_seconds
-    FROM pg_catalog.pg_stat_activity
-    WHERE state is distinct from 'idle' AND query not like 'autovacuum:%'
+	SELECT                                                              
+    COUNT(*) as transactions,
+    MAX(EXTRACT(EPOCH FROM clock_timestamp() - pg_stat_activity.xact_start)) AS oldest_timestamp_seconds
+FROM pg_catalog.pg_stat_activity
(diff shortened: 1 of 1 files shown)


alphabet5 (approved):
I ran into the same issue, and had the same solution. This works as expected. (I'm not a maintainer, so I'm not sure what approving does, but LGTM.)

alphabet5:
Tagging @Sticksman and @SuperQ because you guys worked on the original pr and I'm not really sure what the process is here. Is there anything I can do to help get this into a release?

alphabet5:
I modified the query to exclude a few more things. `idle` was not matching `idle in transaction` queries, and there are a few other common long running queries that are not impactful. I'm not sure if this is useful for everyone though.

```go
	longRunningTransactionsQuery = `
	SELECT
			COUNT(*) as transactions,
			MAX(EXTRACT(EPOCH FROM clock_timestamp() - pg_stat_activity.xact_start)) AS oldest_timestamp_seconds
	FROM pg_catalog.pg_stat_activity
	WHERE state NOT IN ('idle', 'idle in transaction')
			AND query NOT LIKE 'autovacuum:%'
			AND NOT (query ~* '^VACUUM( ANALYZE)?(?! FULL)')
			AND pg_stat_activity.xact_start IS NOT NULL
			AND NOT (query ~* '^REFRESH MATERIALIZED VIEW CONCURRENTLY')
			AND NOT (query ~* '^BEGIN ISOLATION LEVEL READ UNCOMMITTED');
	`
```

roock:
Would be great to have this fixes somewhen, I already reported it in April .

sysadmind (approved):
I'm not sure that it's clear that the metric should be the time that the oldest transaction has been running or if it is supposed to be the timestamp (epoch) of the start of the oldest transaction. Either way it was previously always returning the current time from what I can tell which is not useful to anyone.

PR #1093, opened 2024-11-15
Add connections by client collector

### LINKED ISSUE #969: Add collector for pg_connections
## Proposal
**We had a query to count all connections by machine, which we lost due to deprecation of the queries file. I am sure others would profit from this as well**

```
pg_connections_by_machine:
  query: "select count(*) as count, client_hostname from pg_stat_activity where client_hostname is not null group by client_hostname ;"
  master: true
  metrics:
    - client_hostname:
        usage: "LABEL"
        description: "Name of connecting client"
    - count:
        usage: "GAUGE"
        description: "Number of Connections"
```
If there is another easy solution for this which I am missing right now please tell me.

Regards,
Dennis

Sticksman:
Fixes #969

### DIFF
--- a/collector/pg_connections_by_client.go
+++ b/collector/pg_connections_by_client.go
@@ -68,28 +68,28 @@ func (c PGConnectionByClientCollector) Update(ctx context.Context, instance *ins
 		return err
 	}
 	defer rows.Close()
-	var clientCount sql.NullInt64
-	var clientName sql.NullString
+	var count sql.NullInt64
+	var name sql.NullString
 
 	for rows.Next() {
-		if err := rows.Scan(&clientCount, &clientName); err != nil {
+		if err := rows.Scan(&count, &name); err != nil {
 			return err
 		}
 
-		if !clientName.Valid {
+		if !name.Valid {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


Sticksman:
👓 @SuperQ

collector/pg_connections_by_client.go
```diff
@@ -0,0 +1,100 @@
...
+			"count",
+		),
+		"Number of clients",
+		[]string{"name"}, nil,
+	)
+
+	pgConnectionsByClientQuery = `
+	SELECT
+		count(*) as count,
+		client_hostname
```
sysadmind, line 54: This seems like a bad field to use for this data. According to the docs, this is the reverse dns of the IP. I would prefer to see this use the IP. 

https://www.postgresql.org/docs/17/monitoring-stats.html#MONITORING-PG-STAT-ACTIVITY-VIEW

SuperQ: I agree, I would prefer this to be `client_addr`.

```suggestion
		client_addr
```

SuperQ (changes requested):

PR #1104, opened 2024-12-22
Fix missing dsn sanitization for logging

### LINKED ISSUE #1042: Obfuscate passwords in DSN logs
## Proposal
**Use case. Why is this important?**
We don't want to log our database passwords in our logs. Add a feature to remove passwords when logging out DSN. If there's other auth methods for exporter, maybe it's sufficient to document that using DSN with password will include the password in the log.

sysadmind:
This log line was not sanitized previously which could result in logging sensitive information. I have scanned the rest of the files and I don't see anywhere else that DSN is used in a log line without this filter.

Resolves #1042


sysadmind:
Bump @SuperQ

bboreham (approved):

SuperQ (approved):

PR #1106, opened 2024-12-26
feat: add wait/backend to pg_stat_activity

fgalind1:
This PR adds backend_type, wait_event_type and wait_event labels to the pg_stat_activity metrics. These are very helpful when debugging and trying to understand what kind of queries are stacked in a waiting state and/or where those queries are coming from


ybilyk (approved):

ybilyk (approved):

ybilyk (approved):

sysadmind (approved):
I think there's a chance this could increase cardinality, but I believe that is something we could address if we need to.

PR #1072, opened 2024-10-07
Checkpoint related columns in PG 17 have been moved from pg_stat_bgwriter to pg_stat_checkpointer

n-rodriguez:
Fix https://github.com/prometheus-community/postgres_exporter/issues/1060

See: 
* https://www.dbi-services.com/blog/postgresql-17-new-catalog-view-pg_stat_checkpointer/
* https://www.postgresql.org/docs/16/monitoring-stats.html#MONITORING-PG-STAT-BGWRITER-VIEW
* https://www.postgresql.org/docs/17/monitoring-stats.html#MONITORING-PG-STAT-BGWRITER-VIEW

Thank you!

PR #1112, opened 2025-01-14
Skip pg_stat_checkpointer collector if pg<17

n-rodriguez:
Hello,

Here a fix for https://github.com/prometheus-community/postgres_exporter/pull/1072#issuecomment-2503906073 where the collector fails on pg<17 when we use the exporter with the multi-target pattern

Thank you!

### DIFF
--- a/collector/pg_stat_checkpointer.go
+++ b/collector/pg_stat_checkpointer.go
@@ -110,8 +110,8 @@ var (
 func (c PGStatCheckpointerCollector) Update(ctx context.Context, instance *instance, ch chan<- prometheus.Metric) error {
 	db := instance.getDB()
 
-	before17 := instance.version.Compare(semver.MustParse("17.0.0"))
-	if before17 < 0 {
+	before17 := instance.version.LT(semver.MustParse("17.0.0"))
+	if before17 {
 		c.log.Warn("pg_stat_checkpointer collector is not available on PostgreSQL < 17.0.0, skipping")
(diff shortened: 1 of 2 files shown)


collector/pg_stat_checkpointer.go
```diff
@@ -104,8 +107,15 @@ var (
 	FROM pg_stat_checkpointer;`
 )
 
-func (PGStatCheckpointerCollector) Update(ctx context.Context, instance *instance, ch chan<- prometheus.Metric) error {
+func (c PGStatCheckpointerCollector) Update(ctx context.Context, instance *instance, ch chan<- prometheus.Metric) error {
 	db := instance.getDB()
+
+	before17 := instance.version.LT(semver.MustParse("17.0.0"))
```
sysadmind, line 113: This line is causing the tests to fail because the instance version is not defined. See https://github.com/prometheus-community/postgres_exporter/blob/master/collector/pg_stat_statements_test.go#L33

sysadmind (changes requested):

collector/pg_stat_checkpointer.go
n-rodriguez: ping @michael-todorovic thank you!

n-rodriguez: @sysadmind it should be good now. thank you!

sysadmind (approved):

PR #1116, opened 2025-01-16
Prevent password from being printed to logs

PaulTorm:
Remove Password from logs when printing dsn.


sysadmind:
This was resolved in #1104

PR #1139, opened 2025-03-19
Add datname label to pg_statio_user_indexes.

aleszeleny:
Add `datname` label to `pg_statio_user_indexes` metrics similar to `pg_statio_user_tables` where the `datname` label already exists.
This allows distinguishing indexes with the same fully qualified name (schema name and index name) within different databases running in a PostgreSQL instance.

### DIFF
--- a/collector/pg_statio_user_indexes.go
+++ b/collector/pg_statio_user_indexes.go
@@ -38,18 +38,19 @@ var (
 	statioUserIndexesIdxBlksRead = prometheus.NewDesc(
 		prometheus.BuildFQName(namespace, statioUserIndexesSubsystem, "idx_blks_read_total"),
 		"Number of disk blocks read from this index",
-		[]string{"schemaname", "relname", "indexrelname"},
+		[]string{"datname", "schemaname", "relname", "indexrelname"},
 		prometheus.Labels{},
 	)
 	statioUserIndexesIdxBlksHit = prometheus.NewDesc(
 		prometheus.BuildFQName(namespace, statioUserIndexesSubsystem, "idx_blks_hit_total"),
 		"Number of buffer hits in this index",
-		[]string{"schemaname", "relname", "indexrelname"},
+		[]string{"datname", "schemaname", "relname", "indexrelname"},
 		prometheus.Labels{},
 	)
 
 	statioUserIndexesQuery = `
 	SELECT
+		current_database() AS datname,
 		schemaname,
 		relname,
 		indexrelname,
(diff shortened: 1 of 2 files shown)


ArthurSens:
Hey @aleszeleny, sorry for the delay here. I'm wondering if this change is really necessary... 

Scraping multiple databases now uses the multi-target pattern, and with that pattern, you can easily add the database name label through relabeling rules in the Prometheus scrape config. 

Scraping multiple databases with the single-target pattern is now deprecated, so I'm not sure we want to keep building on it. 

I'm happy to hear more opinions though, cc @sysadmind

aleszeleny:
Hey @ArthurSens, thanks for the comment/explanation.
My motivation (at the time of writing the proposal ) was to fix a gap of a single metric missing the database name label.
My point of view is: If other metrics pg_statio... keep providing the database name label, then for consistency I'd vote for adding the label here as well, despite the Prometheus relabeling feature (or drop it from other metrics -- consistency, but it might break many users' use cases... so probably no viable option).
As you mention, a workaround exists, so it is not a must-have feature/fix, but dealing with a single metrics other way than the majority of others does not seem elegant to me.

Anyway, thanks for the exporter; it's amazing.
Aleš

ArthurSens:
I totally see your point! I believe this behavior was inherited when the exporter was migrated to the prometheus organization. Personally, I'd be in favor of deprecating and eventually removing the database name from the existing metrics instead of adding them everywhere

aleszeleny:
Hey @ArthurSens,

I didn't want to start a flame; therefore: If the change is not acceptable during the evolved project strategy, feel free to close/cancel it.

Personally, I prefer any monitoring system to provide a complete data set regardless of possible post-processing instead of tweaking it to just one (majority) use case. If someone wants use the exporter other way than by scraping the data by Prometheus, missig datname lablel will be an issue. So I'd keep the label on all metrics with scope of the current database and add it if missing somewehere, but it's just my preference.

aleszeleny:
Hey @ArthurSens,

I've found a gotcha (unless there already is a solution I've missed :-) ) - for multi-target monitoring, the instance metrics without `datname` label are scraped from each target database within a single instance, producing a significant volume of duplicate values the timeseries has to deal with. My initial idea collecting all configured metrics only from the "postgres" target database and form other "application" databases drop all metrics missing datname label is not viable due to lack of the `datname` label.

Might it'll help deciding whether or not to accept adding the, from my poit missing, datname lablel.

I've found #1326 so if implemented can be a better solution.

Thanks Ales

ArthurSens:
Hi @aleszeleny, yeah... after reading related discussions in other PRs, it feels like categorizing collectors into cluster/database categories is a better solution.

I'm closing this one so we can focus on the other PRs to solve this problem, I hope you don't mind!

PR #1148, opened 2025-04-05
Deprecated options removed from the README

scottishwidow:
Removed deprecated options related to PG_EXPORTER_EXTEND_QUERY_PATH from README-RDS.
 
PG_EXPORTER_DISABLE_SETTINGS_METRICS and PG_EXPORTER_DISABLE_DEFAULT_METRICS values changed to `false`, because 50% of needed metrics will not be displayed otherwise.


scottishwidow:
@wrouesnel

PR #1149, opened 2025-04-09
Record table only size bytes as well in addition to the total size bytes

Sticksman:
pg_total_relation_size gets the size of the relation + indexes and toast, however we need to know just the size of the table itself. You can get this info from pg_table_size. Should be safe to add, this function was added back in 2010.

### DIFF
--- a/collector/pg_stat_user_tables.go
+++ b/collector/pg_stat_user_tables.go
@@ -157,7 +157,7 @@ var (
 		prometheus.Labels{},
 	)
 	statUserTablesOnlySize = prometheus.NewDesc(
-		prometheus.BuildFQName(namespace, userTableSubsystem, "only_size_bytes"),
+		prometheus.BuildFQName(namespace, userTableSubsystem, "table_size_bytes"),
 		"Total disk space used by only this table, in bytes",
 		[]string{"datname", "schemaname", "relname"},
 		prometheus.Labels{},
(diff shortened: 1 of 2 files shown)


collector/pg_stat_user_tables.go
```diff
@@ -180,7 +186,8 @@ var (
 		autovacuum_count,
 		analyze_count,
 		autoanalyze_count,
-		pg_total_relation_size(relid) as total_size
+		pg_total_relation_size(relid) as total_size,
+		pg_table_size(relid) as table_only_size
```
SuperQ, line 190: Hmm, I wonder if we should refactor this.

> pg_total_relation_size accepts the OID or name of a table or toast table, and returns the total on-disk space used for that table, including all associated indexes. This function is equivalent to pg_table_size + pg_indexes_size.

Instead of exposing the `pg_total_relation_size`, we expose the two underlying metrics and let the users sum up the values themselves.

```suggestion
		pg_indexes_size(relid) as indexes_size,
		pg_table_size(relid) as table_size
```

collector/pg_stat_user_tables.go
```diff
@@ -156,6 +156,12 @@ var (
 		[]string{"datname", "schemaname", "relname"},
 		prometheus.Labels{},
 	)
+	statUserTablesOnlySize = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, userTableSubsystem, "only_size_bytes"),
```
SuperQ, line 160:
```suggestion
		prometheus.BuildFQName(namespace, userTableSubsystem, "table_size_bytes"),
```

SuperQ (approved):
Nice, thanks!

PR #1158, opened 2025-05-16
Feat: Improve Error Handling for Server.Scrape

BoweFlex:
Previously if any errors were encountered by queryNamespaceMappings, only a count of those errors was returned - making debugging those errors harder than it needs to be.

I'm changing this to immediately return nil if no errors are encountered, and otherwise an error will be formatted with each of the namespaces and what the error was for that namespace.

Additionally, if any errors were encountered during querySettings, they were potentially obscured by queryNamespaceMappings errors. I've added an immediate return if errors are found during querySettings to ensure these errors are visible.

### DIFF
--- a/cmd/postgres_exporter/server.go
+++ b/cmd/postgres_exporter/server.go
@@ -129,7 +129,7 @@ func (s *Server) Scrape(ch chan<- prometheus.Metric, disableSettingsMetrics bool
 	}
 	err = fmt.Errorf("queryNamespaceMappings errors encountered")
 	for namespace, errStr := range errMap {
-		err = fmt.Errorf("%s, %s namespace had the following error: %s", err, namespace, errStr)
+		err = fmt.Errorf("%s, namespace: %s error: %s", err, namespace, errStr)
 	}
 
 	return err
--- a/cmd/postgres_exporter/server.go
+++ b/cmd/postgres_exporter/server.go
@@ -119,6 +119,7 @@ func (s *Server) Scrape(ch chan<- prometheus.Metric, disableSettingsMetrics bool
 	if !disableSettingsMetrics && s.master {
 		if err = querySettings(ch, s); err != nil {
 			err = fmt.Errorf("error retrieving settings: %s", err)
+			return err
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/postgres_exporter/server.go
```diff
@@ -119,12 +119,16 @@ func (s *Server) Scrape(ch chan<- prometheus.Metric, disableSettingsMetrics bool
...
 	}
 
 	errMap := queryNamespaceMappings(ch, s)
-	if len(errMap) > 0 {
-		err = fmt.Errorf("queryNamespaceMappings returned %d errors", len(errMap))
+	if len(errMap) == 0 {
+		return nil
+	}
+	for namespace, errStr := range errMap {
+		err = fmt.Errorf("%s\nqueryNamespaceMappings error for the %s namespace: %s", err, namespace, errStr)
```
sysadmind, line 131: I don't see any other errors that we create in this project that have newlines in the error. I worry that adding an error with newlines could break formatting for parsers that our users depend on. I would prefer either a different delimiter, or logging each of these errors here separately however in the latter case, we also need to handle what to return.

sysadmind (changes requested):

cmd/postgres_exporter/server.go
BoweFlex: That makes sense to me. What about a comma separated list of namespaces with their associated errors, i.e.

```text
queryNamespaceMappings errors encountered! %s namespace had the following error: %s, ...
```

sysadmind: I think that sounds reasonable to me.

BoweFlex: Updated as discussed.

sysadmind (approved):
LGTM

cmd/postgres_exporter/server.go
```diff
@@ -119,12 +119,17 @@ func (s *Server) Scrape(ch chan<- prometheus.Metric, disableSettingsMetrics bool
...
 
 	errMap := queryNamespaceMappings(ch, s)
-	if len(errMap) > 0 {
-		err = fmt.Errorf("queryNamespaceMappings returned %d errors", len(errMap))
+	if len(errMap) == 0 {
+		return nil
+	}
+	err = fmt.Errorf("queryNamespaceMappings errors encountered")
+	for namespace, errStr := range errMap {
+		err = fmt.Errorf("%s, %s namespace had the following error: %s", err, namespace, errStr)
```
SuperQ, line 132: Minor nit to make it a little simpler to read (and shorter).

```suggestion
		err = fmt.Errorf("%s, namespace: %s error: %s", err, namespace, errStr)
```

PR #1162, opened 2025-05-29
Add max, mean, and stddev timing metrics to stat_statements collector

irfanhabib:
### Motivation

  The current pg_stat_statements collector only exports total execution time, which makes it difficult to understand query performance patterns. These new metrics allow users to:

  - Identify queries with high execution time variance
  - Detect performance regressions in individual query runs
  - Better understand query performance distribution beyond just totals
  - Set more informed alerting thresholds based on typical vs. maximum execution times

### Changes

  - Added three new Prometheus metric descriptors for max, mean, and standard deviation timing
  - Updated all SQL query templates to include the appropriate columns for different PostgreSQL versions:
    - PostgreSQL < 13: max_time, mean_time, stddev_time
    - PostgreSQL 13+: max_exec_time, mean_exec_time, stddev_exec_time
  - Modified the Update function to scan and emit the new metrics as gauge values
  - Updated test cases to handle the additional columns

### Compatibility

  - Maintains backward compatibility with existing metrics
  - Supports all PostgreSQL versions currently supported by the exporter
  - Uses appropriate column names based on PostgreSQL version detection

### DIFF
--- a/collector/pg_stat_statements.go
+++ b/collector/pg_stat_statements.go
@@ -94,6 +94,24 @@ var (
 		[]string{"user", "datname", "queryid"},
 		prometheus.Labels{},
 	)
+	statStatementsMaxSeconds = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, statStatementsSubsystem, "max_seconds"),
+		"Maximum time spent in a single execution of the statement, in seconds",
+		[]string{"user", "datname", "queryid"},
+		prometheus.Labels{},
+	)
+	statStatementsMeanSeconds = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, statStatementsSubsystem, "mean_seconds"),
+		"Mean time spent per execution of the statement, in seconds",
+		[]string{"user", "datname", "queryid"},
+		prometheus.Labels{},
+	)
+	statStatementsStddevSeconds = prometheus.NewDesc(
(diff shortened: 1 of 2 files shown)


collector/pg_stat_statements.go
```diff
@@ -94,6 +94,24 @@ var (
...
 		prometheus.Labels{},
 	)
+	statStatementsMaxSeconds = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, statStatementsSubsystem, "max_seconds"),
+		"Maximum time spent in a single execution of the statement, in seconds",
+		[]string{"user", "datname", "queryid"},
+		prometheus.Labels{},
+	)
+	statStatementsMeanSeconds = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, statStatementsSubsystem, "mean_seconds"),
```
ArthurSens, line 104: couldn't this be achieved with 
```
postgres_stat_statements_seconds_total
/
postgres_stat_statements_calls_total
```

already? I'd avoid adding more metrics if PromQL can already do it. Folks can use recording rules if they really want a new metric specifically for mean

ArthurSens (commented):
Hi, sorry for the super delayed review here. The idea looks good, but the code has changed a lot in the last few years, so a rebase will be massive amount of work 😅 

I'll close this PR, but please feel free to open a new PR if you're still willing to contribute!

I'm also adding one comment in case you want to open said PR

PR #1165, opened 2025-06-03
Add a collector for `pg_buffercache_summary`.

sfc-gh-pnuttall:
* Add a collector for `pg_buffercache_summary()`
* Requires PostgreSQL >= 16.

### DIFF
--- a/collector/pg_buffercache.go
+++ b/collector/pg_buffercache.go
@@ -1,0 +1,143 @@
+// Copyright 2023 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


collector/pg_buffercache.go
```diff
@@ -0,0 +1,143 @@
...
+import (
+	"context"
+	"database/sql"
+	"log/slog"
+
+	"github.com/blang/semver/v4"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const buffercacheSubsystem = "buffercache"
```
SuperQ, line 25: Since this is technically using `pg_buffercache_summary()`, should we make this `buffercache_summary`?

This would allow us to have both a summary and a full `pg_buffercache` view collector.

collector/pg_buffercache_summary.go
```diff
@@ -0,0 +1,143 @@
...
+			buffers_unused,
+			buffers_dirty,
+			buffers_pinned,
+			usagecount_avg
+		FROM
+		  pg_buffercache_summary()
+		`
+)
+
+func gaugeInt32(m sql.NullInt32, desc *prometheus.Desc, ch chan<- prometheus.Metric) {
```
SuperQ, line 92: @sysadmind, any objection to helper functions like this?

sysadmind: I have no objection to helper function. My preference however is for them to do the conversion and return that value. I don't think I see value in passing in the Desc and Metric chan here.

PR #1166, opened 2025-06-10
[pg_stat_statements] Strip special characters from query

ussrlongbow:
Same queries from Windows and Linux clients to PostgreSQL may appear as different texts in `pg_stat_statements.query` having same query ID, because of different line breaks, which blocks proper matching `query` text to metrics with `queryid` label in Prometheus.
Example:
```
# Series from two replicas, one used by Windows client, another by Linux
$ curl -s 'http://prometheus.local:9090/api/v1/series?match[]=pg_stat_statements_query_id%7Bqueryid=%220000000000%22%7D' | jq .data.[].query 
"SELECT\r\n    "Username"\r\nFROM\r\n    \"Users\"\r\nWHERE\r\n        \"UserID\" = $1"
"SELECT\n    "Username"\nFROM\n    \"Users\"\nWHERE\n        \"UserID\" = $1"
```

This pull request adds option `--[no-]collector.stat_statements.strip_special` to remove such symbols from query output.

Authors:
@ussrlongbow
@freehck

### DIFF
--- a/collector/pg_stat_statements.go
+++ b/collector/pg_stat_statements.go
@@ -27,8 +27,9 @@ import (
 const statStatementsSubsystem = "stat_statements"
 
 var (
-	includeQueryFlag    *bool = nil
-	statementLengthFlag *uint = nil
+	includeQueryFlag    		*bool = nil
+	statementStripSpecialFlag	*bool = nil
+	statementLengthFlag 		*uint = nil
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


collector/pg_stat_statements.go
```diff
@@ -42,6 +43,11 @@ func init() {
 		"Enable selecting statement query together with queryId. (default: disabled)").
 		Default(fmt.Sprintf("%v", defaultDisabled)).
 		Bool()
+	statementStripSpecialFlag = kingpin.Flag(
+		fmt.Sprint(collectorFlagPrefix, statStatementsSubsystem, ".strip_special"),
```
ArthurSens, line 47: `strip_special` is quite generic when all we do here is strip white spaces.

I'm also wondering if this should be configurable at all; maybe the fix should be applied automatically based on `GOOS`?

collector/pg_stat_statements.go
```diff
@@ -105,6 +113,7 @@ var (
 
 const (
 	pgStatStatementQuerySelect = `LEFT(pg_stat_statements.query, %d) as query,`
+	pgStatStatementQuerySelectStrip = `LEFT(regexp_replace(pg_stat_statements.query, '\s+', ' ', 'g'), %d) as query,`
```
ArthurSens, line 116: `\s+` strips a lot more than just CRLF; it strips multiple spaces, tabs, and new lines. Should we use a more restrictive regex?

ArthurSens (changes requested):
Hey, sorry for the very delayed review here. We'll need to rebase this PR to continue 😬

PR #1168, opened 2025-06-18
Add a collector for `pg_proctab`.

sfc-gh-pnuttall:
Docs: https://github.com/markwkm/pg_proctab/tree/main

This collector is useful when:
 * You have access to postgres, but not the underlying machine the postgres server is running on.
 * The `pg_proctab` extension is installed.

We use this in AWS RDS and GCP CloudSQL.

### DIFF
--- a/collector/pg_proctab.go
+++ b/collector/pg_proctab.go
@@ -1,3 +1,16 @@
+// Copyright 2025 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package collector
 
 import (
--- a/collector/pg_proctab.go
+++ b/collector/pg_proctab.go
@@ -1,0 +1,265 @@
+package collector
+
+import (
+	"context"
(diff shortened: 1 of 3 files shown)


collector/pg_proctab.go
```diff
@@ -0,0 +1,278 @@
...
+		log: config.logger,
+	}, nil
+}
+
+var (
+	pgMemusedDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(
+			namespace,
+			proctabSubsystem,
+			"memused",
```
ArthurSens, line 45: Following [Prometheus naming conventions](https://prometheus.io/docs/practices/naming/), the unit of the metric should be part of its name as a suffix. Can we include `_bytes`?

This apply to all other memory metrics

collector/pg_proctab.go
```diff
@@ -0,0 +1,278 @@
...
+	for rows.Next() {
+		if err := rows.Scan(&memused, &memfree, &memshared, &membuffers, &memcached, &swapused); err != nil {
+			return err
+		}
+		emitMemMetric(memused, pgMemusedDesc, ch)
+		emitMemMetric(memfree, pgMemfreeDesc, ch)
+		emitMemMetric(memshared, pgMemsharedDesc, ch)
+		emitMemMetric(membuffers, pgMembuffersDesc, ch)
+		emitMemMetric(memcached, pgMemcachedDesc, ch)
+		emitMemMetric(swapused, pgSwapusedDesc, ch)
```
ArthurSens, line 225: lets move this to after we query the CPU/load metrics. If CPU/load collection fails, this collector will report a failure while providing partial data

ArthurSens (commented):
Hey, sorry for the delayed review here! The codebase has been heavily refactored a few months back, so we might need to do a heavy rebase or start the work fresh from a new PR 😓 

I've added comments below. Also, do we want to include load5, load15, swapfree and swapcached?

PR #1169, opened 2025-06-18
Add a Collector for `pg_stat_io`.

sfc-gh-pnuttall:
Docs: https://www.postgresql.org/docs/current/monitoring-stats.html#MONITORING-PG-STAT-IO-VIEW

We use this in production. One problem is that the stats for a vacuum are reported when the vacuum ends, not continuously.

### DIFF
--- a/cmd/postgres_exporter/server.go
+++ b/cmd/postgres_exporter/server.go
@@ -119,12 +119,17 @@ func (s *Server) Scrape(ch chan<- prometheus.Metric, disableSettingsMetrics bool
 	if !disableSettingsMetrics && s.master {
 		if err = querySettings(ch, s); err != nil {
 			err = fmt.Errorf("error retrieving settings: %s", err)
+			return err
 		}
 	}
 
 	errMap := queryNamespaceMappings(ch, s)
-	if len(errMap) > 0 {
-		err = fmt.Errorf("queryNamespaceMappings returned %d errors", len(errMap))
+	if len(errMap) == 0 {
+		return nil
+	}
+	err = fmt.Errorf("queryNamespaceMappings errors encountered")
+	for namespace, errStr := range errMap {
+		err = fmt.Errorf("%s, namespace: %s error: %s", err, namespace, errStr)
 	}
 
 	return err
--- a/collector/pg_stat_io.go
+++ b/collector/pg_stat_io.go
@@ -1,4 +1,4 @@
-// Copyright 2023 The Prometheus Authors
+// Copyright The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


collector/pg_stat_io.go
```diff
@@ -0,0 +1,325 @@
...
+		}
+
+		if !backendType.Valid || !object.Valid || !PGContext.Valid {
+			continue
+		}
+
+		readsMetric := 0.0
+		if reads.Valid {
+			readsMetric = float64(reads.Int64)
+		}
```
ArthurSens, line 195: This makes a NULL return value appear as 0. I think the right way of doing things would be not to expose the metric at all if there's a NULL return 🤔

ArthurSens: We're doing this in several other metrics too, let's fix all of them

cmd/postgres_exporter/server.go
ArthurSens, line 132: looks unrelated, could we keep it out of this PR?

collector/pg_stat_io.go
```diff
@@ -0,0 +1,325 @@
...
+var (
+	statIOReadsTotal = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, statIOSubsystem, "reads_total"),
+		"Number of read operations, each of the size specified in op_bytes.",
+		[]string{"backend_type", "object", "context"},
+		prometheus.Labels{},
+	)
+	statIOReadTimeTotal = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, statIOSubsystem, "read_time_total"),
+		"Time spent in read operations in milliseconds (if track_io_timing is enabled, otherwise zero)",
```
ArthurSens, line 50: [Prometheus recommends](https://prometheus.io/docs/practices/naming/) the usage of base units, and that the unit is present in the metric name. Could we update all metrics here to follow the conventions?

ArthurSens (commented):
Hi 👋 , sorry for the delayed review! Unfortunately the codebase has been heavily refactored a few months back, so we'll need a heavy rebase or start a fresh new PR 😬 

#1351 added functionality to preperly handle BIGINT colums, we probably want to use these here.

Also, reading docs[[1](https://www.postgresql.org/docs/18/release-18.html#:~:text=Add%20pg_stat_io%20columns,has%20been%20removed.)][[2](https://www.postgresql.org/docs/17/monitoring-stats.html#MONITORING-PG-STAT-IO-VIEW)][[3](https://www.postgresql.org/docs/18/monitoring-stats.html#MONITORING-PG-STAT-IO-VIEW)], it seems like Postgres 18+ usees different fields for this. Could we have version-aware queries (add respective tests)?

PR #1192, opened 2025-09-11
Allow configuring a default and override per-collector timeout

hwreschnig-utiq:
This handles a case where, particularly, the user stat collectors can get blocked behind an exclusive lock preventing the scrape of other metrics from completing before Prometheus gives up; therefore losing all metrics and not being able to see that an exclusive lock has been taken for too long.


cmd/postgres_exporter/probe.go
ArthurSens, line 61: Can we delete this comment since the PR implements a timeout? :)

collector/probe.go
```diff
@@ -86,8 +94,10 @@ func (pc *ProbeCollector) Collect(ch chan<- prometheus.Metric) {
 	wg.Add(len(pc.collectors))
 	for name, c := range pc.collectors {
 		go func(name string, c Collector) {
-			execute(context.TODO(), name, c, pc.instance, ch, pc.logger)
-			wg.Done()
+			ctx, cancel := pc.timeouts.Context(context.TODO(), name)
```
ArthurSens, line 97: Ideally, we don't use a context.TODO(), but use the context from the request.

During a Prometheus scrape, Prometheus sends the scrape timeout as a header; using it would be ideal.

ArthurSens: Thinking a bit more, building on the header is probably better than a configurable timeout 🤔

ArthurSens (commented):
Sounds like a vry good idea. I've added a few comments below!

Also, there's related discussion happening at https://github.com/prometheus/client_golang/issues/1538. We don't need to wait for new interfaces, but we need to think carefully how to migrate what we're building here to whatever comes out from client_golang, as we'd rather to keep this logic there :)

hwreschnig-utiq:
This has diverged so far from the current code after dbb1987 it would need to be rewritten anyway. It looks like the request-scoped probe collector and the long process-scoped collector have been unified so it is now even more difficult than it was before to route a request context into the collection process.

PR #1197, opened 2025-09-24
Fix #1185 bug

sysadmind:
This semicolon breaks the query

/cc @kmoppel


SuperQ (approved):

cristiangreco:
@sysadmind @SuperQ are we good merging this? I could go ahead with https://github.com/prometheus-community/postgres_exporter/pull/1196 then if you're ok

sysadmind:
It's very unfortunate that the tests didn't catch it. I think we need to take a look at how we do testing in this project and see if we can improve.

cristiangreco:
> That's 15 at the moment - due to be upgraded to 17 next weekend as it happens.

Ok I can reproduce the issue locally, will open a PR later.

PR #1198, opened 2025-09-25
Fix swapped `flushedLsn` and `receiveStartTli` for `wal_receiver` collector

spantaleev:
In `pgStatWalReceiverQueryTemplate`, the order of the columns (when `hasFlushedLSN == true`) is:

- ...
- `receive_start_lsn`
- `flushed_lsn`
- `receive_start_tli`
- ...

However, columns were scanned in this order:

- ...
- `receive_start_lsn` -> `receiveStartLsn`
- `receive_start_tli` -> `flushedLsn` (!)
- `flushed_lsn` -> `receiveStartTli` (!)
- ...

This incorrect hydration of variables also manifests as swapped values for the `pg_stat_wal_receiver_flushed_lsn` and `pg_stat_wal_receiver_receive_start_tli` metrics.

This seems to be a bug that has existed since the initial implementation:

- 2d7e15275147ce7dfb7b7dc6f959d60e106c11a1
- https://github.com/prometheus-community/postgres_exporter/pull/844

In this patch, I'm:

- fixing the `.Scan()`, so that it hydrates variables in the correct order

- adjusting the order in which metrics are pushed out to the channel, to follow the order we consume them in (.., `receive_start_lsn`, `flushed_lsn`, `receive_start_tli`, ..)

- adjusting the walreceiver tests, to follow the new order (which matches .`Scan()`)

- fixing a small identation issue in `pgStatWalReceiverQueryTemplate`


spantaleev:
I've fixed the sign-off. Sorry for missing that the first time around!

cristiangreco (approved):

PR #1203, opened 2025-09-29
Update Go

SuperQ:
* Update minimum supported Go to 1.24.0.
* Update Go build to 1.25.x.
* Update PostgreSQL testing versions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ executors:
   # This must match .promu.yml.
   golang:
     docker:
-    - image: cimg/go:1.24
+    - image: cimg/go:1.25
 
 jobs:
   test:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -57,13 +57,11 @@ workflows:
...
-            - circleci/postgres:13
-            - cimg/postgres:14.9
-            - cimg/postgres:15.4
-            - cimg/postgres:16.0
-            - cimg/postgres:17.0
+            - cimg/postgres:13.22
+            - cimg/postgres:14.19
+            - cimg/postgres:15.14
+            - cimg/postgres:16.10
+            - cimg/postgres:17.6
```
cristiangreco, line 64: Are we dropping support for pg 11 / 12 or is this just to stop using the circleci images?

SuperQ: 11/12 are EoL and unsupported by upstream.

We can continue to keep some support in the exporter, but we don't need to waste a lot of CI time on them.

cristiangreco: Thanks for the clarification

Issue #1066, opened 2024-09-08
longRunningTransactionsQuery counts all running queries


grzn:
https://github.com/prometheus-community/postgres_exporter/blob/98f75c7e7ea3a02b974cfeda736cf7dbc091b18e/collector/pg_long_running_transactions.go#L52C2-L52C30

Its missing a condition on the duration of the query, as seen in https://github.com/prometheus-community/postgres_exporter/pull/819/files, e.g.

```
AND (now() - xact_start) > '1 minutes'::interval 
```

![CleanShot 2024-09-09 at 01 15 42](https://github.com/user-attachments/assets/01236073-0074-4e7f-a0fb-17de70b4c26f)

PR #1210, opened 2025-10-14
Fix long_running_transactions query (#1066)

moreinhardt:
Previously the query counted all transactions.

Without this, the metric doesn't really make any sense.

----

Tests still pass. The related test actually mocks the return value of the query so it doesn't test the query itself.

### DIFF
--- a/collector/pg_long_running_transactions.go
+++ b/collector/pg_long_running_transactions.go
@@ -54,6 +54,7 @@ var (
     MAX(EXTRACT(EPOCH FROM clock_timestamp() - pg_stat_activity.xact_start)) AS oldest_timestamp_seconds
 FROM pg_catalog.pg_stat_activity
 WHERE state IS DISTINCT FROM 'idle'
+AND (now() - pg_stat_activity.xact_start) > '1 minutes'::interval
 AND query NOT LIKE 'autovacuum:%'
 AND pg_stat_activity.xact_start IS NOT NULL
 AND pid <> pg_backend_pid();


collector/pg_long_running_transactions.go
ArthurSens: I believe we'll need to turn this variable into NullFloat64, since the new query can now return NULL

ArthurSens (changes requested):
Hello hello, sorry for the delay here!

I believe this is a good improvement, but some work needs to happen before we can merge :) 

It sounds a bit unclear why 1min, if a long running transaction runs for more than 1min then the query would return NULL. I guess this needs at least to be configurable.

Also, could we add tests that verify we don't error when the query return is null?

ArthurSens:
> Not sure how I can make the circleci checks pass? There's some error about a pubkey?

We migrated to GitHub Actions a few months ago, so a simple rebase will do the trick :)

moreinhardt:
> It sounds a bit unclear why 1min, if a long running transaction runs for more than 1min then the query would return NULL. I guess this needs at least to be configurable.

I don't see how my changes _added_ the possible return value of NULL. The query already had filters before; now it just has one more.

I don't know this library well enough but why would an empty set of rows be NULL?

Regarding 1min: I just think it's a reasonable default. The way I have experienced prometheus exporters is that they often use good defaults. IMO making this configurable is a feature that can be added later and should not stop this bug fix PR from being merged. Without my fix this collector literally does not do what it says.


> Also, could we add tests that verify we don't error when the query return is null?

Sorry, I don't have time for that and don't see it related to my changes, as explained above.

moreinhardt:
> > Not sure how I can make the circleci checks pass? There's some error about a pubkey?
> 
> We migrated to GitHub Actions a few months ago, so a simple rebase will do the trick :)

Done.

ArthurSens:
> > It sounds a bit unclear why 1min, if a long running transaction runs for more than 1min then the query would return NULL. I guess this needs at least to be configurable.
> 
> I don't see how my changes _added_ the possible return value of NULL. The query already had filters before; now it just has one more.
> 
> I don't know this library well enough but why would an empty set of rows be NULL?
>
> Regarding 1min: I just think it's a reasonable default. The way I have experienced prometheus exporters is that they often use good defaults. IMO making this configurable is a feature that can be added later and should not stop this bug fix PR from being merged. Without my fix this collector literally does not do what it says.
> 
> > Also, could we add tests that verify we don't error when the query return is null?
> 
> Sorry, I don't have time for that and don't see it related to my changes, as explained above.

Gotcha, thanks a lot for the time spent here, and sorry we couldn't review your PR in a reasonable time. We're currently restructuring the team and hopefully in the future PRs to prometheus exporters will get faster reviews :) 

That said, we can't accept bug fixes without regression tests. I'll close this PR if you don't intend to keep working on it. Feel free to reopen it in case you find some time!

moreinhardt:
> > > It sounds a bit unclear why 1min, if a long running transaction runs for more than 1min then the query would return NULL. I guess this needs at least to be configurable.
> > 
> > 
> > I don't see how my changes _added_ the possible return value of NULL. The query already had filters before; now it just has one more.
> > I don't know this library well enough but why would an empty set of rows be NULL?
> > Regarding 1min: I just think it's a reasonable default. The way I have experienced prometheus exporters is that they often use good defaults. IMO making this configurable is a feature that can be added later and should not stop this bug fix PR from being merged. Without my fix this collector literally does not do what it says.
> > > Also, could we add tests that verify we don't error when the query return is null?
> > 
> > 
> > Sorry, I don't have time for that and don't see it related to my changes, as explained above.
> 
> Gotcha, thanks a lot for the time spent here, and sorry we couldn't review your PR in a reasonable time. We're currently restructuring the team and hopefully in the future PRs to prometheus exporters will get faster reviews :)
> 
> That said, we can't accept bug fixes without regression tests. I'll close this PR if you don't intend to keep working on it. Feel free to reopen it in case you find some time!

Sorry, but this response and the way this PR is just closed feels really uninviting.

I now had a look. The NULL issue was fixed in 778660147e6b6af297005f2c2f51d8fce45380e5 which was included in this PR by the rebase I did today. It was added after my original commit.

Adding a test for the actual thing I fixed would mean adding an integration test with an actual postgres instance.

collector/pg_long_running_transactions.go
moreinhardt: Already fixed in 778660147e6b6af297005f2c2f51d8fce45380e5

ArthurSens:
I've cherry-picked your commit, made it configurable, and added the integration tests in #1379

moreinhardt:
nice! thank you!

PR #1218, opened 2025-11-05
Add COALESCE with -1, to pg_stat_archiver query

aa-popkov:
If coalesce is missing, null may be returned, and then NaN is returned to the prometheus metric. This is incorrect for the gauge type.


ArthurSens:
Hey @aa-popkov, sorry for the delay. This part of the codebase was refactored a few weeks/months ago and, if I'm not mistaken, the problem was already fixed then.

I'll close this PR, but thanks for the suggestion!

Issue #1228, opened 2025-12-09
In case of database lock or low performance prometheus-exporer accumulates connections to the database


pierresouchay:
**What did you do?**

1. Have postgres-exporter running
2. Have a kind of lock to the database -> some requests of postgres-exporter are locked (for instance, on a WAL saturated instance, a big DELETE locking the stats)
3. Have a prometheus server fetching the data

**What did you expect to see?**

1. one or a few connections connection opened to the server
2. Some errors of scrapping maybe

**What did you see instead? Under which circumstances?**

1. On the database, a huge perf issue arise (likely due to dbt)
2. postgres-exporter accumulates postgresql queries stuck
3. Since we did not cap enough the user, it accumulates more than 50% of available connections (in an example, 53 open connections (fair enough, I should have used a user with less privileges, will do), avoiding other clients to connect

I would expect the connections to be canceled at some point (for instance when the prometheus query times out.

Example:

<img width="729" height="763" alt="Image" src="https://github.com/user-attachments/assets/6417a6be-f249-4d83-9ceb-00edc32bebb6" />

We have 53 queries in flight with the user...

**Environment**

* System information:

In the container:
```
uname -srm
Linux 5.15.0-161-generic x86_64
```

* postgres_exporter version:

Using docker image `prometheuscommunity/postgres-exporter:v0.17.1`:

```
postgres_exporter --version
postgres_exporter, version 0.17.1 (branch: HEAD, revision: 1e574cf4fd2a75a8a707d424eafcaa0b88cb7af4)
  build user:       root@b0ba132eddb0
  build date:       20250226-13:53:15
  go version:       go1.23.6
  platform:         linux/amd64
  tags:             unknown
```
* postgres_exporter flags:

None set, using the default ones only

* PostgreSQL version: 17.1

* Logs:
```
insert logs relevant to the issue here
```

pierresouchay:
@sysadmind @SuperQ Fix available in #1229

PR #1229, opened 2025-12-11
Ensure collection of stats succeed in a maximum duration to avoid exhausting PG connections

pierresouchay:
**Context**: As explained in https://github.com/prometheus-community/postgres_exporter/issues/1228, when database is very slow to answer, every call to prometheus exporter might consume a new connection and possibly consume all available connections

**Solution**: Ensure a Context with a specified Timeout is specified, so the connection will end if duration of collection is too long

Will fix https://github.com/prometheus-community/postgres_exporter/issues/1228

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,9 @@ This will build the docker image as `prometheuscommunity/postgres_exporter:${bra
 * `[no-]collector.stat_bgwriter`
   Enable the `stat_bgwriter` collector (default: enabled).
 
+* `[no-]collector.stat_checkpointer`
+  Enable the `stat_checkpointer` collector (default: disabled).
+
 * `[no-]collector.stat_database`
   Enable the `stat_database` collector (default: enabled).
 
--- a/Makefile.common
+++ b/Makefile.common
@@ -112,7 +112,7 @@ common-all: precheck style check_license lint yamllint unused build test
 .PHONY: common-style
 common-style:
 	@echo ">> checking code style"
-	@fmtRes=$$($(GOFMT) -d $$(find . -path ./vendor -prune -o -name '*.go' -print)); \
+	@fmtRes=$$($(GOFMT) -d $$(git ls-files '*.go' ':!:vendor/*' || find . -path ./vendor -prune -o -name '*.go' -print)); \
 	if [ -n "$${fmtRes}" ]; then \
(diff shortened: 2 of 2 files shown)


collector/collector.go
```diff
@@ -158,6 +159,17 @@ func NewPostgresCollector(logger *slog.Logger, excludeDatabases []string, dsn st
 	return p, nil
 }
 
+func CollectionTimeout(s string) Option {
```
sysadmind: I really like options having a "with" prefix.
```suggestion
func WithCollectionTimeout(s string) Option {
```

collector/pg_database_test.go
```diff
@@ -15,13 +15,80 @@ package collector
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/DATA-DOG/go-sqlmock"
 	"github.com/prometheus/client_golang/prometheus"
+
```
sysadmind: I'm not sure what this extra space is for.  I'm surprised that ci isn't complaining. We can remove it. 
```suggestion
```

collector/collector.go
```diff
@@ -166,8 +178,6 @@ func (p PostgresCollector) Describe(ch chan<- *prometheus.Desc) {
 
 // Collect implements the prometheus.Collector interface.
 func (p PostgresCollector) Collect(ch chan<- prometheus.Metric) {
-	ctx := context.TODO()
```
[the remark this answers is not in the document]
pierresouchay: Yes, that's was my initial intention, but did not find a clean simple way to link the http request there. I had to move the context into `collectFromConnection()` instead to make testig simple

README.md
```diff
@@ -255,6 +255,13 @@ The following environment variables configure the exporter:
 * `DATA_SOURCE_PASS_FILE`
   The same as above but reads the password from a file.
 
+* `PG_EXPORTER_COLLECTION_TIMEOUT`
```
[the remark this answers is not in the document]
pierresouchay: I documented it, and better, I now display an error message if value is lower than 1ms (because it likely does not make sense)

collector/collector.go
pierresouchay: done

pierresouchay:
@sysadmind I think I addressed all of your concerns, thank you very much for the review

PR #1231, opened 2025-12-12
Update GitHub Actions to run all CI

sysadmind:
Bonus points - this runs much faster than in circle

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,103 +1,33 @@
 ---
+# Prometheus has switched to GitHub action.
+# Circle CI is not disabled repository-wise so that previous pull requests
+# continue working.
+# This file does not generate any CircleCI workflow.
+
 version: 2.1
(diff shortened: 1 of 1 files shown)


queries.yaml
```diff
@@ -1,2 +1,2 @@
 # Adding queries to this file is deprecated
-# Example queries have been transformed into collectors.
\ No newline at end of file
+# Example queries have been transformed into collectors.
```
sysadmind: This is just a formatting fix that CI won't pass without

.circleci/config.yml
```diff
@@ -1,103 +1,33 @@
 ---
-version: 2.1
+# Prometheus has switched to GitHub action.
```
sysadmind: Let me know if we should just delete this file entirely. I think we need it short term while we get PRs merged or rebased.

PR #1232, opened 2025-12-15
feat: allow excluding databases or users names in `stat_statements`

cristiangreco:
Add a couple of of options to the `stat_statements` collector that allow excluding certain database names or user names from metrics.

This is useful if you don't care about certain preconfigured databases with admin activity.

### DIFF
--- a/collector/pg_stat_statements.go
+++ b/collector/pg_stat_statements.go
@@ -173,7 +173,7 @@ const (
 			WITHIN GROUP (ORDER BY total_time)
 			FROM pg_stat_statements
 		)
-		%s
+		%s %s
 	ORDER BY seconds_total DESC
 	LIMIT %s;`
 
@@ -196,7 +196,7 @@ const (
 			WITHIN GROUP (ORDER BY total_exec_time)
 			FROM pg_stat_statements
 		)
-		%s
+		%s %s
 	ORDER BY seconds_total DESC
 	LIMIT %s;`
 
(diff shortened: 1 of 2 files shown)


collector/pg_stat_statements.go
```diff
@@ -319,3 +361,28 @@ func (c PGStatStatementsCollector) Update(ctx context.Context, instance *instanc
 	}
 	return nil
 }
+
```
sysadmind: Most of these 2 funcs are the same. Maybe we should consolidate to 1 function that just does the escaping and returns the slices.

collector/pg_stat_statements.go
```diff
@@ -198,11 +238,13 @@ func (c PGStatStatementsCollector) Update(ctx context.Context, instance *instanc
...
 		querySelect = fmt.Sprintf(pgStatStatementQuerySelect, c.statementLength)
 	}
+	databaseFilter := c.buildExcludedDatabasesClause()
+	userFilter := c.buildExcludedUsersClause()
 	statementLimit := defaultStatementLimit
 	if c.statementLimit > 0 {
 		statementLimit = fmt.Sprintf("%d", c.statementLimit)
 	}
-	query := fmt.Sprintf(queryTemplate, querySelect, statementLimit)
+	query := fmt.Sprintf(queryTemplate, querySelect, databaseFilter+userFilter, statementLimit)
```
sysadmind: Why put them both in the same parameter? It makes it less obvious. I think it would be better to have separate parameters.

PR #1246, opened 2026-01-25
feat: Add granted label pg_locks collector

### LINKED ISSUE #657: Add "granted" to pg_locks_count as a label
## Proposal

The current `pg_locks_count` metric doesn't differentiate between granted and not granted locks. To learn about the current congestion in the database, it would be good to divide the number of locks between granted and not-yet-granted locks. 

Link to the table definition in the PostgreSQL docs: https://www.postgresql.org/docs/current/view-pg-locks.html

For now, I'll go with a custom query, still I'd assume this feature would be helpful for other as well.

Starting with PostgreSQL 14 there is also the column `waitstart` that would allow to calculate the accumulated waiting time for all not-yet-granted locks. I assume that would be a different metric.

eugercek:
Hi,

This PR adds `granted` label to `pg_locks_count` metric.

Implements and Closes #657.

## Implementation Choices:

**Postgres version compatibility**: I believe `pg_locks.granted` exists since postgres started to using Git. This is [the commit](https://github.com/postgres/postgres/commit/580fb7fb41414da359133a0342505dd3c9a7756a) that `src/backend/catalog/system_views.sql` file is added and `pg_locks.granted `exists at that time too. Nov 11 2003. Also in [7.3.0 release notes](https://www.postgresql.org/docs/release/7.3.0/ ) `pg_locks` is mentioned. So It'll be available since 7.3.0, because of these I did not use `instance.version.GE` like check, it's too old.

**Metric format**: I did not touch current format it'll be cartesian product, `value=0` will be exists. Only added `granted` label.

**How bool is implemented**: Since we already use data point's value for count of the current data point, like in below metric we know that there are 9 locks for `granted="true",mode="accesssharelock"`, I can't use the metric's value like [`pg_replication_slot_slot_is_active`](https://github.com/prometheus-community/postgres_exporter/blob/9e42fc0145a9d780d824d961afb7b70867558424/collector/pg_replication_slot.go#L58-L65), so used a new label called `granted`.
```promql
pg_locks_count{datname="postgres",granted="true",mode="accesssharelock"} 9`
```

Because we use count value I couldn't use [this standard](https://github.com/prometheus-community/postgres_exporter/blob/9e42fc0145a9d780d824d961afb7b70867558424/collector/pg_replication_slot.go#L58-L65). Found open-telemetry/opentelemetry-dotnet#4822 standard looks like using lowercase `"true" "false"` so used that.

**Cardinality and Metric Count**: Cardinality will be 2, just `granted="true|false"`. But metric count will be doubled because of the added 2 new dimension.

## Tests

**Unit Test**:

```bash
go test ./collector -run TestPGLocksCollector -v
=== RUN   TestPGLocksCollector

  Metrics comparison ✔


1 total assertion

--- PASS: TestPGLocksCollector (0.00s)
PASS
ok      github.com/prometheus-community/postgres_exporter/collector     (cached)
```

**Manual test**

Ran below:
```
CREATE DATABASE foo;
-- \c foo
CREATE TABLE foo(name text);

begin;
LOCK TABLE foo IN EXCLUSIVE MODE; 

-- In another session
begin;
LOCK TABLE foo IN EXCLUSIVE MODE; 
```

` curl -s http://localhost:9187/metrics  | grep pg_locks | grep foo`

<details>

<summary>Output</summary>

```promql
pg_locks_count{datname="foo",granted="false",mode="accessexclusivelock"} 0
pg_locks_count{datname="foo",granted="false",mode="accesssharelock"} 0
pg_locks_count{datname="foo",granted="false",mode="exclusivelock"} 1
pg_locks_count{datname="foo",granted="false",mode="rowexclusivelock"} 0
pg_locks_count{datname="foo",granted="false",mode="rowsharelock"} 0
pg_locks_count{datname="foo",granted="false",mode="sharelock"} 0
pg_locks_count{datname="foo",granted="false",mode="sharerowexclusivelock"} 0
pg_locks_count{datname="foo",granted="false",mode="shareupdateexclusivelock"} 0
pg_locks_count{datname="foo",granted="false",mode="sireadlock"} 0
pg_locks_count{datname="foo",granted="true",mode="accessexclusivelock"} 0
pg_locks_count{datname="foo",granted="true",mode="accesssharelock"} 0
pg_locks_count{datname="foo",granted="true",mode="exclusivelock"} 1
pg_locks_count{datname="foo",granted="true",mode="rowexclusivelock"} 0
pg_locks_count{datname="foo",granted="true",mode="rowsharelock"} 0
pg_locks_count{datname="foo",granted="true",mode="sharelock"} 0
pg_locks_count{datname="foo",granted="true",mode="sharerowexclusivelock"} 0
pg_locks_count{datname="foo",granted="true",mode="shareupdateexclusivelock"} 0
pg_locks_count{datname="foo",granted="true",mode="sireadlock"} 0
```
</details>

### DIFF
--- a/collector/pg_locks.go
+++ b/collector/pg_locks.go
@@ -45,43 +45,52 @@ var (
 			"count",
 		),
 		"Number of locks",
-		[]string{"datname", "mode"}, nil,
+		[]string{"datname", "mode", "granted"}, nil,
 	)
 
 	pgLocksQuery = `
-		SELECT 
+		SELECT
 		  pg_database.datname as datname,
-		  tmp.mode as mode,
-		  COALESCE(count, 0) as count 
-		FROM 
+		  tmp_mode.mode as mode,
+		  tmp_granted.granted as granted,
+		  COALESCE(count, 0) as count
+		FROM
 		  (
-		    VALUES 
-		      ('accesssharelock'), 
-		      ('rowsharelock'), 
-		      ('rowexclusivelock'), 
-		      ('shareupdateexclusivelock'), 
-		      ('sharelock'), 
-		      ('sharerowexclusivelock'), 
-		      ('exclusivelock'), 
-		      ('accessexclusivelock'), 
+		    VALUES
+		      ('accesssharelock'),
+		      ('rowsharelock'),
+		      ('rowexclusivelock'),
+		      ('shareupdateexclusivelock'),
+		      ('sharelock'),
+		      ('sharerowexclusivelock'),
+		      ('exclusivelock'),
+		      ('accessexclusivelock'),
 		      ('sireadlock')
-		  ) AS tmp(mode)
-		  CROSS JOIN pg_database 
+		  ) AS tmp_mode(mode)
+		  CROSS JOIN (
+		    VALUES
+		      ('true'),
+		      ('false')
+		  ) AS tmp_granted(granted)
+		  CROSS JOIN pg_database
 		  LEFT JOIN (
-		    SELECT 
-		      database, 
-		      lower(mode) AS mode, 
-		      count(*) AS count 
-		    FROM 
-		      pg_locks 
-		    WHERE 
-		      database IS NOT NULL 
-		    GROUP BY 
-		      database, 
-		      lower(mode)
-		  ) AS tmp2 ON tmp.mode = tmp2.mode 
-		  and pg_database.oid = tmp2.database 
-		ORDER BY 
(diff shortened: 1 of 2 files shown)


ArthurSens (commented):
Adding a new label to an existing metric is a breaking change, so the least we can do here is a very big warning to the changelog 😬 

If I understand it correctly, this will also break our mixin alerts and dashboards, could you take a look at those?

PR #1248, opened 2026-01-27
chore(Dashboard): Use default timepicker

ChristianTackeGSI:
Just use the system's default timepicker instead of hardcoding it.


ArthurSens (approved):
LGTM, thanks!

ChristianTackeGSI:
Thanks! :-)

PR #1263, opened 2026-02-18
feat: do not load config file when path is empty

### LINKED ISSUE #794: Error opening config file postgres_exporter.yml in logs
**What did you do?**
Install/Upgrade to 0.12.0

Tried to add `--config.file=''` and `--config.file='false'` to disable the config file feature but it keeps trying to read the config file. 
Also a default of 'postgres_exporter.yml' doesn't really specify where this file should be located.

No errors in the logs

Error in the logs
```
caller=main.go:86 level=error msg="Error loading config" err="Error opening config file \"postgres_exporter.yml\": open postgres_exporter.yml: no such file or directory"
```
**Environment**

* System information:

`Linux 3.10.0-1160.88.1.el7.x86_64 x86_64`

* postgres_exporter version:

```
# postgres_exporter --version
postgres_exporter, version 0.12.0 (branch: HEAD, revision: f9a1edbbc6d629663aa658378e1d989293e6e488)
  build user:       root@463311dfdfb6
  build date:       20230321-23:52:03
  go version:       go1.20.2
  platform:         linux/amd64
  tags:             netgo static_build
```
* postgres_exporter flags:

Contents of Defaults file:
```
### Managed by SaltStack
DATA_SOURCE_NAME=postgresql://postgres_exporter:********@127.0.0.1:5432/postgres?sslmode=disable
PG_EXPORTER_WEB_LISTEN_ADDRESS=:9187
PG_EXPORTER_DISABLE_DEFAULT_METRICS=false
PG_EXPORTER_DISABLE_SETTINGS_METRICS=false
PG_EXPORTER_EXTEND_QUERY_PATH=/usr/share/prometheus/queries.yaml
PG_EXPORTER_AUTO_DISOVER_DATABASES=true
PG_EXPORTER_EXCLUDE_DATABASES=template0,template1

POSTGRES_EXPORTER_OPTS="--extend.query-path=/usr/share/prometheus/queries.yaml --web.listen-address=:9187"
### EOF ###
```

* PostgreSQL version:

not relevant

* Logs:

litetex:
Fixes #794

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -77,16 +77,18 @@ func (ch *Handler) ReloadConfig(f string, logger *slog.Logger) error {
 		}
 	}()
 
-	yamlReader, err := os.Open(f)
-	if err != nil {
-		return fmt.Errorf("error opening config file %q: %s", f, err)
-	}
-	defer yamlReader.Close()
-	decoder := yaml.NewDecoder(yamlReader)
-	decoder.KnownFields(true)
+	if f != "" {
+		yamlReader, err := os.Open(f)
+		if err != nil {
+			return fmt.Errorf("error opening config file %q: %s", f, err)
+		}
+		defer yamlReader.Close()
+		decoder := yaml.NewDecoder(yamlReader)
+		decoder.KnownFields(true)
 
-	if err = decoder.Decode(config); err != nil {
-		return fmt.Errorf("error parsing config file %q: %s", f, err)
+		if err = decoder.Decode(config); err != nil {
+			return fmt.Errorf("error parsing config file %q: %s", f, err)
+		}
 	}
 
 	ch.Lock()


sysadmind (changes requested):
This input is almost guaranteed to never be empty. It has a default in the flag. What are you trying to achieve with this change because I don't think this achieves it?

vnea:
Will it be merged?

ArthurSens:
Hey, sorry for the delay here! 

We recently did a pretty big refactor, so you'll need to rebase and redo your change. 

I do believe it is a good change, though, so I'll try to help you get this merged! A single target Postgres exporter shouldn't need a config file and we shouldn't provide warnings for that

litetex:
Hey so this is getting ridiculous:
* All I did was adding a single quick and dirty `if string is not empty`
* I was told that the PR is not valid - which I answered that it was - and then I never heard any feedback again
* now half a year later I'm getting notifications if I can redo the entire PR because everything was refactored. But the underlying problem was (apparently) overlooked AGAIN.

I'm currently not even using the project nor do I have the repo anymore.

Well how about you do the change yourself because if you already have the repo setup that's an effort of like 1min... oh and I also don't have the risk of wasting my time and waiting half a year again.

ArthurSens:
I'm sorry you feel this way @litetex. We're all volunteers here, doing this work unpaid in our free time. The Prometheus organization has hundreds of exporters and fewer than 5 maintainers across them, so unfortunately reviews do get delayed a bit.

I will open a PR with the fix you suggested. Thank you for suggesting the change!

PR #1274, opened 2026-03-05
fix(collector): handle Aurora's unsupported pg_last_xact_replay_times…

### LINKED ISSUE #1273: Bug: Replication collector crashes on Aurora PostgreSQL
## Summary

The replication collector (`pg_replication`) fails on every scrape cycle when
running against Amazon Aurora PostgreSQL. This causes the entire collection
cycle to abort with an error, producing no metrics for that scrape interval.

## Root Cause

Aurora PostgreSQL does not support `pg_last_xact_replay_timestamp()`. When the
replication collector executes its primary query, which calls this function,
Aurora returns:

```
ERROR: pg_last_xact_replay_timestamp() is currently not supported for Aurora
```

Because the collector treats any scan error as fatal, this error propagates up
and terminates the scrape. No replication metrics (including `is_replica`, which
Aurora *can* answer) are emitted.

## Impact

- Any postgres_exporter deployment targeting Aurora PostgreSQL instances loses
  all replication metrics on every scrape.
- The error appears in exporter logs on every scrape interval, producing noise.
- `pg_replication_is_replica`, which is useful for distinguishing writer from
  reader instances, is never emitted even though Aurora supports
  `pg_is_in_recovery()`.

## Expected Behavior

The collector should degrade gracefully on Aurora: emit `NaN` for the
time-based metrics that rely on `pg_last_xact_replay_timestamp()`, and still
emit `pg_replication_is_replica` using the supported `pg_is_in_recovery()`
function.

## Environment

- postgres_exporter v0.19.x
- Amazon Aurora PostgreSQL (any version)

dannotripp:
# fix(collector): handle Aurora's unsupported pg_last_xact_replay_timestamp

Fixes #1273

Aurora PostgreSQL does not support `pg_last_xact_replay_timestamp()`, causing
the replication collector to abort every scrape with a fatal error on Aurora
instances.

This change detects the Aurora-specific `feature_not_supported` error (Postgres
error class `0A`) and falls back gracefully: `pg_replication_is_replica` is
still reported via `pg_is_in_recovery()`, while the time-based metrics emit
`NaN` to signal they are unavailable.

A new test `TestPgReplicationCollectorAurora` covers the fallback path.

### DIFF
--- a/collector/pg_replication.go
+++ b/collector/pg_replication.go
@@ -15,7 +15,12 @@ package collector
 
 import (
 	"context"
+	"database/sql"
+	"errors"
+	"math"
+	"strings"
 
+	"github.com/lib/pq"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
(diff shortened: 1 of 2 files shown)


leonardobenedet (approved):

collector/pg_replication.go
```diff
@@ -72,32 +77,71 @@ var (
...
+			row2 := db.QueryRowContext(ctx, pgReplicationIsReplicaQuery)
+			if err2 := row2.Scan(&isReplica); err2 != nil {
+				isReplica = 0
+			}
+		} else {
+			return err
+		}
 	}
+
+	lagValue := math.NaN()
```
sysadmind, line 125: Why NaN over just not emitting the metric? The latter is common in other collectors when the database value is NULL.

PR #1275, opened 2026-03-23
Add a collector for pending restarts

mpldr:
Produces something along the lines of:
```
# HELP pg_catalog_pending_restart_info PostgreSQL configuration parameters that are pending a server restart to take effect.
# TYPE pg_catalog_pending_restart_info gauge
pg_catalog_pending_restart_info{pending_value="300",setting="max_connections"} 1
```

To test:
```sql
ALTER SYSTEM SET max_connections = '200'; SELECT pg_reload_conf();
```

### DIFF
--- a/collector/pg_catalog.go
+++ b/collector/pg_catalog.go
@@ -1,0 +1,88 @@
+// Copyright 2022 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


MarcWort:
Hi, I just saw this PR, and it is exactly what I need, thank you for creating this @mpldr.

Is there anything blocking this from being reviewed or merged? I’d be happy to help test or make updates if needed.

MarcWort:
@SuperQ Could you please take a look at this? It would be much easier if it was merged, rather than me having to maintain a local patch and build for my systems.

ArthurSens (commented):
Hi, sorry for the delayed review here! The codebase has been heavily refactored a few months back, so the PR will need a big rebase 😬 

Also, we need to have proper test coverage before accepting any PRs.

If you don't mind, could you clarify ehat is the intended use case for the `pending_value` label? `pg_settings.setting` contains the currently active value, not the value awaiting restart. Are you trying to expose only which settings require a restart, or also the configured value that will take effect afterward?

ArthurSens:
We could probably re-use the existing settings collector instead of creating a new one too 🤔

PR #1292, opened 2026-04-20
Add sequence_overflow metrics

alphabet5:
This query was based on

https://www.crunchydata.com/blog/the-integer-at-the-end-of-the-universe-integer-overflow-in-postgres

and exposes metrics for % used for the underlying column type.

This requires usage on all sequences, or `GRANT pg_read_all_data TO postgres_exporter;`

pg_read_all_data is a bit excessive, but granting usage on all sequences won't apply for new sequences that are added, and setting default privileges for schemas won't apply if new schemas are added.

```text
pg_sequence_overflow_column_ratio{column_datatype="bigint", datname="mydb", instance="mydb.example.com", job="postgres-exporter", owned_by="my_table.id", schemaname="public", sequence="my_table_id_seq", sequence_datatype="bigint"}
	8.745220514546834e-10
pg_sequence_overflow_last_value{column_datatype="bigint", datname="mydb", instance="mydb.example.com", job="postgres-exporter", owned_by="my_table.id", region="dal", schemaname="public", sequence="my_table_id_seq", sequence_datatype="bigint"}
	8066042235
pg_sequence_overflow_sequence_ratio{column_datatype="bigint", datname="mydb", instance="mydb.example.com", job="postgres-exporter", owned_by="my_table.id", schemaname="public", sequence="my_table_id_seq", sequence_datatype="bigint"}
	8.745220514546834e-10
```

### DIFF
--- a/collector/pg_sequence_overflow.go
+++ b/collector/pg_sequence_overflow.go
@@ -1,0 +1,214 @@
+// Copyright 2026 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package collector
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"log/slog"
+	"net/url"
+	"slices"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const sequenceOverflowSubsystem = "sequence_overflow"
+
+func init() {
+	registerCollector(sequenceOverflowSubsystem, defaultDisabled, NewPGSequenceOverflowCollector)
+}
+
+type PGSequenceOverflowCollector struct {
+	log               *slog.Logger
+	excludedDatabases []string
+}
+
+func NewPGSequenceOverflowCollector(config collectorConfig) (Collector, error) {
+	return &PGSequenceOverflowCollector{
+		log:               config.logger,
+		excludedDatabases: config.excludeDatabases,
+	}, nil
+}
+
+var (
+	sequenceOverflowLastValue = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, sequenceOverflowSubsystem, "last_value"),
+		"Last value consumed from the sequence.",
+		[]string{"datname", "schemaname", "sequence", "sequence_datatype", "owned_by", "column_datatype"},
+		prometheus.Labels{},
+	)
+	sequenceOverflowSequenceRatio = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, sequenceOverflowSubsystem, "sequence_ratio"),
+		"Ratio of the last sequence value to the maximum value of the sequence datatype (0–1).",
+		[]string{"datname", "schemaname", "sequence", "sequence_datatype", "owned_by", "column_datatype"},
+		prometheus.Labels{},
+	)
+	sequenceOverflowColumnRatio = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, sequenceOverflowSubsystem, "column_ratio"),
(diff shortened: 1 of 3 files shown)


collector/pg_sequence_overflow.go
```diff
@@ -0,0 +1,214 @@
...
+// It is called by the Prometheus registry when collecting metrics.
+// Note: pg_sequence_last_value requires USAGE privilege on each sequence,
+// or pg_read_all_data. Without it the function returns NULL, which COALESCE converts
+// to 0, causing all metrics to show 0 with no visible error.
+func (c *PGSequenceOverflowCollector) Update(ctx context.Context, instance *instance, ch chan<- prometheus.Metric) error {
+	db := instance.getDB()
+
+	// Fetch the list of connectable, non-template databases.
+	rows, err := db.QueryContext(ctx, sequenceOverflowDatabaseQuery)
+	if err != nil {
```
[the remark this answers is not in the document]
sysadmind: I'm not bothered by this one. We only officially test back to version 13. It's also not enabled by default, so a user has to opt in.

collector/pg_sequence_overflow.go
```diff
@@ -0,0 +1,214 @@
...
+JOIN pg_namespace ns ON ns.oid = seqs.relnamespace
+WHERE d.deptype = 'a'
+  AND d.classid = 1259;
+`
+)
+
+// dsnForDatabase returns a new DSN with the database name replaced.
+// It updates both the URL path and the dbname query parameter (if present),
+// since libpq gives precedence to the query parameter over the path.
+func dsnForDatabase(dsn string, datname string) (string, error) {
```
sysadmind, line 99: This functionality is unique to this collector. I think the intention is that these need to be run per database, but I think that other functionality would benefit from that too. I don't think that it should be specific to this collector. It's also not obvious that this collector would make new connections.

collector/pg_sequence_overflow.go
```diff
@@ -0,0 +1,214 @@
...
+    seqs.relname AS sequence,
+    format_type(s.seqtypid, NULL) AS sequence_datatype,
+    CONCAT(tbls.relname, '.', attrs.attname) AS owned_by,
+    format_type(attrs.atttypid, atttypmod) AS column_datatype,
+    ns.nspname AS schemaname,
+    COALESCE(pg_sequence_last_value(seqs.oid::regclass), 0) AS last_sequence_value,
+    COALESCE(CASE format_type(s.seqtypid, NULL)
+        WHEN 'smallint' THEN pg_sequence_last_value(seqs.oid::regclass) / 32767::float
+        WHEN 'integer'  THEN pg_sequence_last_value(seqs.oid::regclass) / 2147483647::float
+        WHEN 'bigint'   THEN pg_sequence_last_value(seqs.oid::regclass) / 9223372036854775807::float
```
sysadmind, line 78: I don't think this exporter needs to calculate these values. That can be done in the Prometheus query. Having the raw value is much more useful for Prometheus. It means you can calculate this value on the fly and you know the absolute value, all with only storing a single time series.

collector/pg_sequence_overflow.go
```diff
@@ -0,0 +1,214 @@
...
+    END, 0) AS sequence_ratio,
+    COALESCE(CASE format_type(attrs.atttypid, NULL)
+        WHEN 'smallint' THEN pg_sequence_last_value(seqs.oid::regclass) / 32767::float
+        WHEN 'integer'  THEN pg_sequence_last_value(seqs.oid::regclass) / 2147483647::float
+        WHEN 'bigint'   THEN pg_sequence_last_value(seqs.oid::regclass) / 9223372036854775807::float
```
sysadmind, line 83: Same here as above - we can calculate this at Prometheus query time.

collector/pg_sequence_overflow.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	// Query sequence metrics from each database individually.
+	for _, datname := range databases {
+		newDSN, err := dsnForDatabase(instance.dsn, datname)
+		if err != nil {
+			c.log.Debug("Skipping database", "datname", datname, "err", err)
+			continue
+		}
+		dbConn, err := sql.Open("postgres", newDSN)
```
sysadmind, line 153: I agree with copilot - this creates a lot of new connections which might not be desirable for users.

sysadmind (changes requested):
The commit is also missing the DCO which is a requirement for this project. The CI details have instructions for how to fix it.

PR #1296, opened 2026-05-04
Remove global auth config reload metrics

ArthurSens:
This change separates the exporter auth configuration from the runtime handler that reloads it and observes reload metrics.

It removes the auth reload metrics from `promauto`-style global registration. Instead, each auth config handler creates and registers its own reload metrics with a caller-provided registerer. 

The exporter binary still uses an auth config handler, so existing reload behavior is preserved. The important change is that auth config data, auth config loading, reload state, and reload metric registration are no longer collapsed into one global-ish abstraction. The exporter binary also continues to use the Global Default Registerer

### DIFF
--- a/cmd/postgres_exporter/main.go
+++ b/cmd/postgres_exporter/main.go
@@ -35,9 +35,7 @@ import (
 )
 
 var (
-	c = config.Handler{
-		Config: &config.Config{},
-	}
+	c = newConfigHandler()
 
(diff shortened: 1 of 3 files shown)


sysadmind (approved):
This looks good to me. We typically try to make metric descriptions package vars, but I'm not opposed to this. I know there are also opinions against those package vars because of the large namespace it creates.

ArthurSens:
I don't really know the process in this repository. Should I be merging the PR myself? Or should I wait for other reviews?

PR #1299, opened 2026-05-05
Config pkg now holds all user-provided configuration knobs

ArthurSens:
This includes things that were previously configurable only through CLI flags. The existing CLI flags now construct the new Config Struct

This PR builds on top of #1297

### DIFF
--- a/cmd/postgres_exporter/main.go
+++ b/cmd/postgres_exporter/main.go
@@ -43,9 +43,11 @@ type legacyFlags struct {
 }
 
 var (
-	c                  = newAuthConfigHandler()
-	cfg                = config.NewConfigWithDefaults()
-	legacyMetricsFlags = legacyFlags{}
+	c                              = newAuthConfigHandler()
+	cfg                            = config.NewConfigWithDefaults()
+	legacyMetricsFlags             = legacyFlags{}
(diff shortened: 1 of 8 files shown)


cmd/postgres_exporter/main.go
```diff
@@ -34,25 +34,68 @@ import (
 	"github.com/prometheus/exporter-toolkit/web/kingpinflag"
 )
 
+type legacyFlags struct {
+	AutoDiscoverDatabases bool
+	UserQueriesPath       string
+	ConstantLabels        string
+	ExcludeDatabases      string
+	IncludeDatabases      string
+}
```
ArthurSens, line 43: My understanding is that we want to remove deprecated CLI flags, so I'm not adding them to the new Config struct 😬

ArthurSens:
Here is an [example PR](https://github.com/prometheus/prometheus-opentelemetry-collector/pull/11/changes#diff-41b735a3ca77a5c5b4f91da664fd2f59be61c3c47ee8be65ec2944beff68a9da) that consumes the `Config` Struct introduced here.

ArthurSens:
closing in favor of https://github.com/prometheus-community/postgres_exporter/pull/1330

PR #1302, opened 2026-05-13
feat: Add probe target allowlist for /probe (SSRF and DNS rebinding hardening)

adeptvin1:
## Summary

Adds an optional **allowlist** for the multi-target **`/probe`** endpoint so operators can restrict which hosts/IPs and ports the exporter may connect to. This mitigates **SSRF-style abuse** of an open `/probe` URL and hardens checks against **DNS rebinding**. When the allowlist is **non-empty**, resolved addresses must match **CIDR** and/or **host:port** entries. **`sslmode=verify-full`** semantics are preserved when the probe DSN is rewritten.

## Motivation

With `/probe?target=…`, anyone who can reach the exporter can ask it to connect to arbitrary Postgres endpoints. In shared or less trusted networks this is risky. An explicit allowlist is a standard control for this pattern (similar to other multi-target exporters).

## Security context (Prometheus model)

The [Prometheus security model](https://prometheus.io/docs/operating/security/) notes that some exporters accept targets via URL parameters, so anyone who can reach the exporter’s HTTP endpoint may be able to make it connect to **arbitrary** endpoints. The multi-target `/probe?target=…` pattern in postgres_exporter falls into that category.

This change does **not** replace network controls, reverse proxies, authentication, or the general guidance not to expose monitoring HTTP to untrusted networks as the primary defense. It adds an **optional**, configuration-side **restriction on where `/probe` may connect**, for operators who want extra control on top of their deployment model.

## Intended use

The feature is **opt-in** and **off by default**: an empty allowlist keeps existing behavior (all targets allowed). It is aimed mainly at deployments where the exporter is reachable from a **broader or less trusted part of a corporate network** (or similar), and operators still use `/probe` instead of a sidecar—**only when they choose** to constrain which Postgres endpoints may be probed through this instance.

## Behavior

- **Flag:** `--probe-target-allowlist`
- **Environment variable:** `PG_EXPORTER_PROBE_TARGET_ALLOWLIST`
- **Format:** comma-separated **`CIDR`** and/or **`host:port`** (for example `10.0.0.0/8`, `db.example.com:5432`)
- **Default (empty):** unchanged behavior — **all** probe targets allowed (backward compatible)

## Implementation notes

- Allowlist validation uses resolved IPs where relevant to reduce **DNS rebinding** issues.
- Probe DSN rewriting keeps **TLS `verify-full`** behavior consistent with the connection path.

## Documentation and tests

- README updated (usage, flags, env var).
- Unit tests in `probe_allowlist_test.go`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ To use the multi-target functionality, send an http request to the endpoint `/pr
 
 To avoid putting sensitive information like username and password in the URL, preconfigured auth modules are supported via the [auth_modules](#auth_modules) section of the config file. auth_modules for DSNs can be used with the `/probe` endpoint by specifying the `?auth_module=foo` http parameter.
 
+For additional hardening, `/probe` targets can be restricted with `--probe-target-allowlist` (or `PG_EXPORTER_PROBE_TARGET_ALLOWLIST`). The allowlist accepts comma-separated `CIDR` and `host:port` entries. By default this setting is empty, which preserves existing behavior and allows all probe targets.
+
 Example Prometheus config:
 ```yaml
 scrape_configs:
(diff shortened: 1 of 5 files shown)


adeptvin1:
Good day!
Do you have any updates?

PR #1303, opened 2026-05-13
Refactor pg_settings into a standalone collector

ArthurSens:


collector/pg_setting.go
```diff
@@ -11,77 +11,87 @@
...
 
-// pgSetting is represents a PostgreSQL runtime variable as returned by the
+// pgSetting represents a PostgreSQL runtime variable as returned by the
 // pg_settings view.
 type pgSetting struct {
 	name, setting, unit, shortDesc, vartype string
 }
 
-func (s *pgSetting) metric(labels prometheus.Labels) (prometheus.Metric, error) {
+func (s *pgSetting) metric() (prometheus.Metric, error) {
```
ArthurSens, line 89: This seems like a breaking change. Previously, it was possible to pass const_labels to pg_settings metrics, but as far as I understand, that's no longer possible with the new collector architecture.

Is that ok?

sysadmind (approved):

collector/pg_setting.go
sysadmind: I think it's okay. We just need to call it out in the changelog.

PR #1304, opened 2026-05-13
Refactor stat_archiver into standalone collector

ArthurSens:


collector/pg_stat_archiver.go
```diff
@@ -0,0 +1,109 @@
...
+
+	var archivedCount, failedCount sql.NullInt64
+	var lastArchiveAge sql.NullFloat64
+
+	if err := row.Scan(&archivedCount, &failedCount, &lastArchiveAge); err != nil {
+		return err
+	}
+
+	archivedCountMetric := 0.0
+	if archivedCount.Valid {
```
sysadmind, line 78: I think we can move sending the metric to the channel into these if statements. I _think_ that a metric of 0 is misleading here. I would rather skip emitting the metric (like a NULL) if we can't get valid data.

collector/pg_stat_archiver.go
```diff
@@ -0,0 +1,109 @@
...
+	"database/sql"
+
+	"github.com/blang/semver/v4"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const statArchiverSubsystem = "stat_archiver"
+
+func init() {
+	registerCollector(statArchiverSubsystem, defaultEnabled, NewPGStatArchiverCollector)
```
sysadmind, line 27: I need to check the rest of the code, but I think we may be mixing hyphens and underscores in these names. We should standardize on one or the other. Underscores probably make the most sense because that is the metric name subsystem.

ArthurSens: I don't see how hyphens are getting mixed up here 🤔. We control the whole metric name, nothing is dynamically generated.

ArthurSens: During a manual test, this is the result I get from the query
```
archived_count=0
failed_count=0
last_archived_time=NULL
last_archive_age=NULL
```

And these are the metrics:
```
# HELP pg_scrape_collector_duration_seconds postgres_exporter: Duration of a collector scrape.
# TYPE pg_scrape_collector_duration_seconds gauge
pg_scrape_collector_duration_seconds{collector="stat_archiver"} 0.014364333
# HELP pg_scrape_collector_success postgres_exporter: Whether a collector succeeded.
# TYPE pg_scrape_collector_success gauge
pg_scrape_collector_success{collector="stat_archiver"} 1
# HELP pg_stat_archiver_archived_count Number of WAL files that have been successfully archived
# TYPE pg_stat_archiver_archived_count counter
pg_stat_archiver_archived_count 0
# HELP pg_stat_archiver_failed_count Number of failed attempts for archiving WAL files
# TYPE pg_stat_archiver_failed_count counter
pg_stat_archiver_failed_count 0
```

sysadmind: I think my comment was a bit ambiguous, my apologies. I didn't mean the rest of your code mixing hyphens and underscores. I mean the whole code base might not be consistent.

sysadmind: I took a quick pass and I don't see any hyphens standing out in the rest of the code base. I think we're good.

sysadmind (approved):

PR #1305, opened 2026-05-13
Refactor stat_activity into standalone collector

ArthurSens:


collector/pg_stat_activity.go
```diff
@@ -0,0 +1,185 @@
...
+			stringValue(state),
+			stringValue(usename),
+			stringValue(applicationName),
+			stringValue(backendType),
+			stringValue(waitEventType),
+			stringValue(waitEvent),
+		}
+
+		countMetric := 0.0
+		if count.Valid {
```
sysadmind, line 154: I think we should also put the metric send inside the if here (and the other PRs). In general I hesitate to use the zero value if we don't know that we actually got data. It can be misleading for alerts where a 0 value might indicate healthy.

ArthurSens:
PR rebased and review addressed :) 

I'll do the same thing for my other PRs regarding the NULL results

sysadmind (approved):

PR #1307, opened 2026-05-13
Refactor stat_replication_slots into standalone collector

ArthurSens:


README.md
```diff
@@ -136,6 +136,9 @@ This will build the docker image as `prometheuscommunity/postgres_exporter:${bra
 * `[no-]collector.replication_slot`
   Enable the `replication_slot` collector (default: enabled).
 
+* `[no-]collector.replication_slots`
+  Enable the `replication_slots` collector (default: enabled).
+
```
ArthurSens: I've just noticed the existence of these two collectors, with the only difference being one singular and one plural. Looks kinda weird to me, but I don't know postgres enough to say if this is correct or incorrect 😅

sysadmind: Looking at the queries between the two, they're the same database (the plural). I think we should combine them. I like the plural name because it matches the table.

ArthurSens:
@sysadmind anything else needed from my side? :)

sysadmind (approved):

egerkke:
Hi,

According to my view, changes that requires the users to adapt their solution are non-backward compatible changes. Renaming for me is a non-backward compatible change since I need to change on what metrics I receive information.
Can you please explain why you haven't increased major version according to the SemVer rules?

Thanks and BR.
Gergo

Issue #611, opened 2022-02-07
No need to run `GRANT postgres_exporter1 TO <MASTER_USER>` in aws RDS?


richzw:
Per doc

> -- If deploying as non-superuser (for example in AWS RDS), uncomment the GRANT
-- line below and replace <MASTER_USER> with your root user.
-- GRANT postgres_exporter1 TO <MASTER_USER>;

However, I create one user for postgres_exporter without `GRANT postgres_exporter1 TO <MASTER_USER>` as below

```sql
CREATE OR REPLACE FUNCTION __tmp_create_user() returns void as $$
BEGIN
  IF NOT EXISTS (
          SELECT
          FROM   pg_catalog.pg_user
          WHERE  usename = 'postgres_exporter1') THEN
    CREATE USER postgres_exporter1;
  END IF;
END;
$$ language plpgsql;

SELECT __tmp_create_user();
DROP FUNCTION __tmp_create_user();

ALTER USER postgres_exporter1 WITH PASSWORD 'EFcPCDftzxXspO3o';
ALTER USER postgres_exporter1 SET SEARCH_PATH TO pg_catalog;

GRANT CONNECT ON DATABASE postgres TO postgres_exporter1;

GRANT pg_monitor to postgres_exporter1;
```

And this user `postgres_exporter` could work well in postgres_exporter.

Should we think this operation `GRANT postgres_exporter1 TO <MASTER_USER>` is useless? Please correct us if something is wrong or anything missing?

PR #1308, opened 2026-05-13
Add Amazon Aurora PostgreSQL support

### LINKED ISSUE #1273: Bug: Replication collector crashes on Aurora PostgreSQL

megative:
## Motivation

Amazon Aurora PostgreSQL is a very common deployment target for this
exporter, but the support today is at best lossy:

- **`replication` collector fails every scrape on Aurora.** Aurora
  rejects `pg_last_xact_replay_timestamp()` with `feature_not_supported`,
  so the collector returns an error, emits no replication metrics for
  that scrape interval, and writes the same error to the exporter log on
  every tick. This is tracked in #1273; a user there reports the bug
  hitting "over three hundred databases".
- **`wal` collector fails the same way.** Aurora does not support
  `pg_ls_waldir()`, so the standard WAL metrics also blow up on every
  scrape with no graceful path.
- **Aurora-specific subsystems are invisible.** Aurora exposes its
  storage replication lag, Optimized Reads cache, per-database
  commit-latency / DML-activity stats, and Global Database state via
  Aurora-only built-in functions (`aurora_replica_status()`,
  `aurora_stat_database()`, `aurora_global_db_status()`, …). None of
  these are observable through the standard `pg_*` views, so operators
  on Aurora typically scrape them via ad-hoc custom queries or stitch in
  CloudWatch metrics by hand.

This exporter already ships `README-RDS.md` and Aurora-related issues
recur in the tracker (#611, #1273), so the user base is real. Putting
Aurora support upstream — behind one opt-in flag, with the existing
non-Aurora behaviour fully preserved — avoids forks duplicating the same
work and gives Aurora operators a single tested integration.

## Summary

* New top-level flag `--aurora.enabled` (env: `PG_EXPORTER_AURORA_ENABLED`,
  default `false`). When set, the exporter runs `SELECT aurora_version()`
  once per instance to detect Aurora and flips the default state of every
  `aurora_*` collector to enabled. Per-collector `--collector.aurora_X` /
  `--no-collector.aurora_X` flags still win.
* 9 new opt-in collectors covering Aurora-specific functions:
  `aurora_replica_status`, `aurora_global_db_status`,
  `aurora_global_db_instance_status`, `aurora_stat_database`,
  `aurora_stat_bgwriter`, `aurora_stat_commit_latency`,
  `aurora_stat_dml_activity`, `aurora_stat_optimized_reads_cache`,
  `aurora_stat_logical_wal_cache`. All disabled by default. On non-Aurora
  servers they short-circuit to `ErrNoData` (no log noise, no extra
  queries).
* Graceful degradation in `replication` and `wal`: when the underlying
  function comes back with PostgreSQL error class `0A`
  (`feature_not_supported`) **and** the message contains `"Aurora"`, the
  collector emits NaN for the time-/WAL-directory-based metrics and still
  emits `pg_replication_is_replica` via `pg_is_in_recovery()`. Fixes
  #1273. The fingerprint is narrow on purpose — non-Aurora users see no
  behaviour change.

## Design notes

* Zero overhead on non-Aurora deployments: when `--aurora.enabled=false`
  (default), the `aurora_version()` probe is skipped entirely and every
  `aurora_*` collector is a no-op. The flag must be opt-in.
* The Aurora probe runs at most once per process: `*auroraProbe` carries
  a `sync.Once` and is shared by every `instance.copy()` made per scrape.
* `--aurora.enabled` is a top-level flag rather than per-collector
  because it controls a probe that affects multiple collectors at once.

## Open questions

Happy to address in-PR — flagging up-front so it doesn't get missed:

* Several Aurora metrics carry their AWS-native units in the name
  (e.g. `aurora_replica_lag_milliseconds`,
  `aurora_stat_commit_latency_microseconds_total`) because they mirror
  the underlying column names. If you'd prefer base units in seconds
  (matching the pattern in `pg_stat_statements`), happy to convert.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go test ./...` all clean.
- [x] Unit tests added for each new collector (`*_test.go`) covering
  happy path with `isAurora=true`, `ErrNoData` on `isAurora=false`,
  empty result set, and `excludeDatabases` filtering where applicable.
- [x] `pg_replication` / `pg_wal` tests cover both the standard path
  and the Aurora-unsupported-function fallback.
- [x] README updated with the new collectors and a note on the
  graceful-degradation behaviour.
- [x] CHANGELOG entry added under `## main / (unreleased)`.

Closes #1273.

This is an AI assisted PR.


sysadmind:
This PR is pretty large as is. I'm in favor of improving our Aurora support and I would like to see this work get merged. I think the best way for that to happen is to split this PR up into smaller, easier to review PRs. I think the new collectors are good candidates to either be solo PRs or maybe small groups that are closely related. For some of the other bug fixes, can you double check open PRs? I think at least one of those bug callouts has an open PR that I gave feedback on recently. I want to give people who already opened PRs the chance to get their work merged before we take a different approach.

megative:
Thanks for the feedback! Will definitely double check the open PRs just in case.

Just for context - we're currently running this version of the exporter across all our environments, including production, so I'm very motivated to get this upstreamed properly.

Also, I would be happy to split it into smaller, more focused PRs - for purposes of future review. I understand there's huge amount of changes but I believe this is a big step forward for exporter functionality since there's no proper aurora support in other exporters as well. I decided to patch it since postgres exporter is a common used solution I'm working with for a long years, and I think it would help community a lot.

Could you please specify what can I exactly do to merge this PR into the upstream?

cmd/postgres_exporter/main.go
```diff
@@ -50,20 +52,13 @@ var (
 	includeDatabases       = kingpin.Flag("include-databases", "A list of databases to include when autoDiscoverDatabases is enabled (DEPRECATED)").Default("").Envar("PG_EXPORTER_INCLUDE_DATABASES").String()
 	metricPrefix           = kingpin.Flag("metric-prefix", "A metric prefix can be used to have non-default (not \"pg\") prefixes for each of the metrics").Default("pg").Envar("PG_EXPORTER_METRIC_PREFIX").String()
 	collectionTimeout      = kingpin.Flag("collection-timeout", "Timeout for collecting the statistics when the database is slow").Default("1m").Envar("PG_EXPORTER_COLLECTION_TIMEOUT").String()
+	auroraEnabled          = kingpin.Flag("aurora.enabled", "Enable Amazon Aurora PostgreSQL support. When set, the exporter detects Aurora on each instance and turns Aurora-specific collectors on by default. Has no effect on non-Aurora instances.").Default("false").Envar("PG_EXPORTER_AURORA_ENABLED").Bool()
```
SuperQ: We generally try and avoid this kind of feature flag. It's preferred to try and detect the version of the database and use appropriate collectors when detected.

See the `version` field on the `instance` struct.

megative:
> This PR is pretty large as is. I'm in favor of improving our Aurora support and I would like to see this work get merged. I think the best way for that to happen is to split this PR up into smaller, easier to review PRs. I think the new collectors are good candidates to either be solo PRs or maybe small groups that are closely related. For some of the other bug fixes, can you double check open PRs? I think at least one of those bug callouts has an open PR that I gave feedback on recently. I want to give people who already opened PRs the chance to get their work merged before we take a different approach.

Checked — #1274 is already covers the `pg_replication` Aurora fix. So I'll drop that part from my work and let it land there.

cmd/postgres_exporter/main.go
megative: @SuperQ 

Done in 57a79ec — `--aurora.enabled` is deleted now (along with the env var and the three `WithX` helpers).

Mirrors the `version` pattern you pointed at: `instance.isAurora` is set in `setup()` via a one-shot `SELECT aurora_version()` (`sync.Once`, so one probe per process), and `aurora_*` collectors check it in `Update()`
the same way others check `version`.

Per-collector `--collector.aurora_X` toggles stay default-disabled (same as `stat_checkpointer`, also version-gated).

Seems like should be OK for now

megative:
@sysadmind — how would you suggest splitting this up? One PR per Aurora collector, or some other grouping?

sysadmind:
I think that once per collector is ideal.

megative:
> I think that once per collector is ideal.

@sysadmind first PR of the series is ready: #1311 (aurora_replica_status + Aurora auto-detection foundation). 
Happy to queue the remaining 8 per-collector PRs right after, or should I wait for #1311 to land first so the architecture pattern can settle?

sysadmind:
@megative I reviewed #1311 and based on the volume of feedback, I would recommend we settle the architecture first to save you having to redo any work.

PR #1311, opened 2026-05-17
Add aurora_replica_status collector

megative:
## Summary

First PR in a series splitting #1308 (Aurora support) into one-PR-per-collector as requested by @sysadmind.

This PR adds:

* **Aurora auto-detection infrastructure on `instance` struct.** Mirrors the existing `version` field pattern (per @SuperQ feedback). On every`setup()`, a one-shot `SELECT aurora_version()` runs via `sync.Once` — exactly one probe per process lifetime. The result is cached in `instance.isAurora` and read by Aurora-specific collectors.

* **`aurora_replica_status` collector.** Exposes the values from Aurora's `aurora_replica_status()` built-in: replica lag (per server), replay latency, and pending read IOs. Disabled by default (`--collector.aurora_replica_status`); on non-Aurora servers it short-circuits to `ErrNoData` cleanly.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go test ./...` clean.
- [x] Tested live against an Aurora PostgreSQL 17.7.0 cluster — collector
  emits metrics; per-process `aurora_version()` probe fires exactly once.
- [x] Tested against vanilla PostgreSQL 17 — collector returns `ErrNoData`
  silently; no error logs.

Follow-up PRs in the series will add the remaining 8 Aurora collectors on top of this infrastructure.


megative:
hi @SuperQ @sysadmind 
Could you guys please take a look? It would be great to deliver Aurora support to upstream.

megative:
@SuperQ @sysadmind guys, sorry for pushing but any luck you have a chance to review this one?

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 ## main / (unreleased)
 
 * [CHANGE] ...
-* [FEATURE] ...
+* [FEATURE] Add `aurora_replica_status` collector for Amazon Aurora PostgreSQL, with Aurora auto-detection on the `instance` struct.
```
ArthurSens, line 4:
```suggestion
* [FEATURE] Add `aurora_replica_status` collector for Amazon Aurora PostgreSQL.
```

The second part of the phrase seems like implementation details; the postgres_exporter end-user doesn't care what an "instance struct" is

collector/aurora_replica_status.go
```diff
@@ -0,0 +1,103 @@
...
+
+type AuroraReplicaStatusCollector struct{}
+
+func NewAuroraReplicaStatusCollector(collectorConfig) (Collector, error) {
+	return &AuroraReplicaStatusCollector{}, nil
+}
+
+var (
+	auroraReplicaLagDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, auroraReplicaSubsystem, "lag_milliseconds"),
```
ArthurSens, line 40: prometheus recommends using [base units](https://prometheus.io/docs/practices/naming/#base-units) for metric units. We have a few examples of other collectors doing this transformation, like pg_setting and pg_stat_statements

collector/aurora_replica_status.go
```diff
@@ -0,0 +1,103 @@
...
+		"Replica lag behind writer in milliseconds (aurora_replica_status.replica_lag_in_msec).",
+		[]string{"server_id"}, nil,
+	)
+	auroraReplayLatencyDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, auroraReplicaSubsystem, "replay_latency_microseconds"),
```
ArthurSens, line 45: same applies here

collector/instance.go
ArthurSens, line 58: asking without thinking too much... could the aurora detection be made here? If I understand it correctly, this runs only once and never again. So if we made the detection here, it would make the sync.Once shenanigans unecessary

ArthurSens: I'm probably missing context too since I'm not a maintainer of this exporter, but I'm wondering why version detection isn't made here either

collector/instance.go
```diff
@@ -17,19 +17,36 @@ import (
...
 	db      *sql.DB
 	version semver.Version
+
+	// isAurora is set by setup() via the shared auroraProbe (one-shot
+	// SELECT aurora_version() per process). Aurora-specific collectors
+	// gate on this field; on non-Aurora servers they short-circuit to
+	// ErrNoData.
+	isAurora    bool
+	auroraProbe *auroraProbe
 }
```
ArthurSens, line 44: why `isAurora` set in both `instance` and `auroraProbe`? Seems like instance has access to two fields that have the same meaning/value

collector/instance.go
```diff
@@ -65,14 +83,36 @@ func (i *instance) setup() error {
 	} else {
 		i.version = version
 	}
+
+	if i.auroraProbe != nil {
+		i.auroraProbe.once.Do(func() {
+			i.auroraProbe.isAurora = detectAurora(i.db)
+		})
+		i.isAurora = i.auroraProbe.isAurora
+	}
```
ArthurSens, line 92: I don't see this sync.Once protection against the query that detects version. I'm not sure if I'm missing something, but looks like an overcomplication....?

ArthurSens (commented):
Disclaimer, I'm not a maintainer of postgres_exporter, I'm just trying to unload a bit of their review load :)

collector/aurora_replica_status.go
sysadmind: agreed. This should be converted.

collector/instance.go
sysadmind: We also have the option to do detection where we do PostgreSQL instance version checks.

collector/instance.go
sysadmind: I don't think we need both. I think it's complicated and doesn't add much value.

collector/aurora_replica_status.go
```diff
@@ -0,0 +1,103 @@
...
+	"context"
+	"database/sql"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// auroraReplicaStatusCollector exposes metrics from Aurora PostgreSQL's
+// aurora_replica_status() function. Disabled by default because the function
+// only exists on Aurora.
+const auroraReplicaSubsystem = "aurora_replica"
```
sysadmind, line 26: I think it's unintuitive for the subsystem and the collector name below to be different. Someone runs `--collector.aurora_replica_status` but the metrics are `aurora_replace_*`. These should match. Based on the table name, they should both be `aurora_replica_status`.

collector/aurora_replica_status_test.go
```diff
@@ -0,0 +1,122 @@
...
+	go func() {
+		defer close(ch)
+		c := AuroraReplicaStatusCollector{}
+		if err := c.Update(context.Background(), inst, ch); err != nil {
+			t.Errorf("Error calling Update: %s", err)
+		}
+	}()
+
+	// writer: 1 metric (pending_read_ios), reader: 3 metrics → 4 total
+	expected := 4
```
sysadmind, line 56: This only validates that we had 4 metrics, not what they were. It's incomplete.

README.md
```diff
@@ -199,6 +199,15 @@ This will build the docker image as `prometheuscommunity/postgres_exporter:${bra
...
   Enable the `xlog_location` collector (default: disabled).
 
+#### Amazon Aurora PostgreSQL collectors
+
+These collectors expose metrics that only exist on Amazon Aurora PostgreSQL.
+They are disabled by default and detect Aurora automatically — on a
+non-Aurora server they silently emit no data.
+
+* `[no-]collector.aurora_replica_status`
+  Replica lag, replay latency, pending read IOs from `aurora_replica_status()`. (default: disabled)
```
sysadmind, line 209: I think calling out each metric here makes it easy to miss when the collector changes.

sysadmind (changes requested):

ArthurSens:
Friendly ping @megative, do you still plan to work on aurora support?

PR #1313, opened 2026-05-17
stat_replication: fix "slot_name does not exist" by joining pg_replication_slots

### LINKED ISSUE #1310: Bug: stat_replication collector crashes ("slot_name" does not exist)
## Summary

The `stat_replication` collector (introduced in #1306) fails every scrape with:

```
pq: column "slot_name" does not exist
```

This affects **every user on `main`** after #1306, on **any** PostgreSQL version. The collector is `defaultEnabled`, so the breakage is automatic.

## Root cause

The query in `collector/pg_stat_replication.go` selects `slot_name` from `pg_stat_replication`:

```sql
SELECT application_name, client_addr::text, state, slot_name, ...
FROM pg_stat_replication
```

But `slot_name` is **not** a column of `pg_stat_replication` on any supported PostgreSQL version. It lives on `pg_replication_slots`.

### Verification

Three independent sources:

1. `\d pg_stat_replication` on `postgres:17.10.0` Docker — 20 columns, no `slot_name`.
2. `\d pg_stat_replication` on `postgres:18` Docker — 20 columns, no `slot_name`.
3. Official docs for [PG 17](https://www.postgresql.org/docs/17/monitoring-stats.html#MONITORING-PG-STAT-REPLICATION-VIEW) and [PG 18](https://www.postgresql.org/docs/18/monitoring-stats.html#MONITORING-PG-STAT-REPLICATION-VIEW) — 20 columns documented, no `slot_name`. The PG 18 docs explicitly note `slot_name` lives on `pg_stat_replication_slots`.

### Why CI did not catch it

Before #1306 the same collector ran via the legacy yaml-driven exporter with a `SELECT * FROM pg_stat_replication` query. The yaml mapping listed `slot_name` as a label, but since the column did not exist, the yaml exporter silently produced an empty label — no error. #1306 promoted `slot_name` into an explicit SQL column list, turning that silent no-op into a hard parse-time error. Unit tests use `sqlmock` and do not validate against the real Postgres schema, so the bug shipped.

## Impact

- Every user on `main` after #1306 loses `pg_stat_replication_*` metrics on every scrape.
- Exporter logs spam the same error every tick.
- Default-enabled collector — no opt-in needed to trigger.
- Affects Aurora and vanilla Postgres equally.

## Proposed solution

`LEFT JOIN pg_replication_slots s ON s.active_pid = pg_stat_replication.pid` and read `s.slot_name` from there. Behavior:

- Replication client using a named slot → `slot_name` populated.
- Replication client without a slot → `slot_name` empty (matches legacy yaml behavior).
- Aurora and other setups without physical replication slots → `slot_name` empty.

PR #1313 implements this.

## Environment

- postgres_exporter `main` (after #1306 merge)
- Confirmed on: vanilla PG 17.10.0 (Docker), vanilla PG 18 (Docker), Amazon Aurora PostgreSQL 17.7.0

megative:
## Summary

Fixes #1310. The `stat_replication` collector currently fails every scrape with `pq: column "slot_name" does not exist` — affects every user on `main` after #1306, not just Aurora (despite the issue title).

The fix is a `LEFT JOIN pg_replication_slots s ON s.active_pid = pg_stat_replication.pid` so that `slot_name` is read from the view where it actually lives, with an empty label when no named slot is in use.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go test ./...` clean.
- [x] Live: vanilla PG 17.10.0 — `pg_scrape_collector_success{collector="stat_replication"} = 1` (was `0`).
- [x] Live: Aurora PG 17.7.0 — `pg_scrape_collector_success{collector="stat_replication"} = 1` (was `0`).

<details>
<summary>Full timeline and root cause</summary>

### Timeline

1. **Before #1306**, `pg_stat_replication` was scraped through the legacy yaml-driven exporter with a `SELECT * FROM pg_stat_replication` query. The yaml mapping listed `slot_name` as a label, but the column does not actually exist on `pg_stat_replication` — so the yaml exporter silently produced an empty label. No error.

2. **`slot_name` has never been a column of `pg_stat_replication`** on any supported PG version. It lives on `pg_replication_slots`. To populate it for an active WAL sender you have to JOIN the two views. The legacy yaml never did. The label was effectively aspirational — described, but always empty. A latent bug sat in `queries.yaml` for years.

3. **#1306 ("Refactor stat_replication into standalone collector")** migrated this query into a Go-native collector. The new SQL was rewritten with an explicit column list, including `slot_name`:

   ```sql
   SELECT application_name, client_addr::text, state, slot_name, ...
   FROM pg_stat_replication
   ```

   That promoted `slot_name` from "yaml label silently missing" to "explicit SQL column that must exist". It does not.

4. **CI on #1306 went green** because unit tests for `stat_replication` use `sqlmock`, which returns whatever columns the test declares — the real Postgres schema is never validated. Integration tests apparently do not exercise this collector. The bug shipped.

5. **After #1306 merged**, `stat_replication` fails on every scrape, on every PG version. Because it is `defaultEnabled`, every user is affected automatically.

### Verification that `slot_name` is not on `pg_stat_replication`

Three independent sources:

1. `\d pg_stat_replication` on `postgres:17.10.0` Docker — 20 columns, no `slot_name`.
2. `\d pg_stat_replication` on `postgres:18` Docker — 20 columns, no `slot_name`.
3. Official docs for [PG 17](https://www.postgresql.org/docs/17/monitoring-stats.html#MONITORING-PG-STAT-REPLICATION-VIEW) and [PG 18](https://www.postgresql.org/docs/18/monitoring-stats.html#MONITORING-PG-STAT-REPLICATION-VIEW) — 20 columns documented, no `slot_name`. The PG 18 docs explicitly note `slot_name` lives on `pg_stat_replication_slots`.

### Fix details

`LEFT JOIN pg_replication_slots s ON s.active_pid = pg_stat_replication.pid` and read `s.slot_name`. Behavior:

* Replication client using a named slot → `slot_name` populated.
* Replication client without a slot → `slot_name` empty (matches legacy yaml behavior).
* Aurora and other setups without physical replication slots → `slot_name` empty.

Applied to both the PG ≥ 10 and pre-10 query paths. `pg_replication_slots` exists since PG 9.4, so the pre-9.4 branch remains affected by the same root cause; that branch is not in CI scope and was already broken before this PR.

</details>

### DIFF
--- a/collector/pg_stat_replication.go
+++ b/collector/pg_stat_replication.go
@@ -55,15 +55,20 @@ var (
 		prometheus.Labels{},
 	)
 
+	// slot_name is not a column of pg_stat_replication; it lives on
+	// pg_replication_slots and is linked via the standby's WAL sender
+	// PID. LEFT JOIN so clients that do not use a named slot still get
+	// a row with an empty slot_name label.
 	statReplicationQuery = `
 			SELECT
 				application_name,
 				client_addr::text,
 				state,
-				slot_name,
+				s.slot_name,
 				(case pg_is_in_recovery() when 't' then pg_wal_lsn_diff(pg_last_wal_receive_lsn(), pg_lsn('0/0'))::float else pg_wal_lsn_diff(pg_current_wal_lsn(), pg_lsn('0/0'))::float end) AS pg_current_wal_lsn_bytes,
 				(case pg_is_in_recovery() when 't' then pg_wal_lsn_diff(pg_last_wal_receive_lsn(), replay_lsn)::float else pg_wal_lsn_diff(pg_current_wal_lsn(), replay_lsn)::float end) AS pg_wal_lsn_diff
 			FROM pg_stat_replication
+			LEFT JOIN pg_replication_slots s ON s.active_pid = pg_stat_replication.pid
 			`
 
 	statReplicationQueryBefore10 = `
@@ -71,7 +76,20 @@ var (
 				application_name,
 				client_addr::text,
 				state,
-				slot_name,
+				s.slot_name,
+				(case pg_is_in_recovery() when 't' then pg_xlog_location_diff(pg_last_xlog_receive_location(), replay_location)::float else pg_xlog_location_diff(pg_current_xlog_location(), replay_location)::float end) AS pg_xlog_location_diff
+			FROM pg_stat_replication
+			LEFT JOIN pg_replication_slots s ON s.active_pid = pg_stat_replication.pid
+			`
+
+	// pg_replication_slots.active_pid only exists on PostgreSQL 9.5+, so
+	// on older versions slot_name cannot be resolved; the label is left
+	// empty (matching the legacy yaml exporter behavior).
+	statReplicationQueryBefore95 = `
+			SELECT
+				application_name,
+				client_addr::text,
+				state,
 				(case pg_is_in_recovery() when 't' then pg_xlog_location_diff(pg_last_xlog_receive_location(), replay_location)::float else pg_xlog_location_diff(pg_current_xlog_location(), replay_location)::float end) AS pg_xlog_location_diff
 			FROM pg_stat_replication
 			`
@@ -81,8 +99,10 @@ func (PGStatReplicationCollector) Update(ctx context.Context, instance *instance
 	switch {
 	case instance.version.GTE(semver.MustParse("10.0.0")):
 		return updateStatReplication(ctx, instance, ch)
-	case instance.version.GTE(semver.MustParse("9.2.0")):
+	case instance.version.GTE(semver.MustParse("9.5.0")):
 		return updateStatReplicationBefore10(ctx, instance, ch)
+	case instance.version.GTE(semver.MustParse("9.2.0")):
+		return updateStatReplicationBefore95(ctx, instance, ch)
 	default:
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


collector/pg_stat_replication.go
```diff
@@ -55,25 +55,31 @@ var (
...
 	statReplicationQueryBefore10 = `
 			SELECT
 				application_name,
 				client_addr::text,
 				state,
-				slot_name,
+				s.slot_name,
 				(case pg_is_in_recovery() when 't' then pg_xlog_location_diff(pg_last_xlog_receive_location(), replay_location)::float else pg_xlog_location_diff(pg_current_xlog_location(), replay_location)::float end) AS pg_xlog_location_diff
 			FROM pg_stat_replication
+			LEFT JOIN pg_replication_slots s ON s.active_pid = pg_stat_replication.pid
```
ArthurSens, line 82: Seems like this query only works for postgres 9.5 and above, because `active_pid` was only added in PG 9.5.x. See https://www.postgresql.org/docs/9.5/view-pg-replication-slots.html and https://www.postgresql.org/docs/9.4/catalog-pg-replication-slots.html

Could you please have a query for 9.4 and below that doesn't include `slot_name`?

ArthurSens (changes requested):
Thanks a lot for the fix, and apologies for not seeing this PR before making the 0.20.0 release.

There's still some small work to be done to make sure this also works with older versions of PG.

Could you also rebase the PR?

collector/pg_stat_replication.go
megative: Hi @ArthurSens !

Rebased on master and added a fallback for 9.2–9.4: those versions use a query without the join, `slot_name` label is just empty there (same as the old yaml behavior).

Verified against real postgres in docker. On 9.4 the join query indeed fails:

```
ERROR:  column s.active_pid does not exist
```

the fallback query runs fine and the collector succeeds:

```
pg_scrape_collector_success{collector="stat_replication"} 1
```

On 17 with a standby streaming through a named slot (`pg_basebackup -R -S test_slot`) the exporter emits:

```
pg_stat_replication_pg_current_wal_lsn_bytes{application_name="walreceiver",client_addr="172.19.0.3/32",slot_name="test_slot",state="streaming"} 1.1758988e+08
pg_stat_replication_pg_wal_lsn_diff{application_name="walreceiver",client_addr="172.19.0.3/32",slot_name="test_slot",state="streaming"} 0
```

9.6 with the join query works too. Unit tests pass, incl. a new one for the pre-9.5 path.

ArthurSens (approved):
Thanks!

PR #1315, opened 2026-05-21
add pg_stat_statements plans

jfcoz:
Add pg_stat_statements metrics:
plans_total: Number of times planned
plans_seconds_total: Total time spent in planning the statement

It needs the PostgreSQL setting: pg_stat_statements.track_planning

### DIFF
--- a/collector/pg_stat_statements.go
+++ b/collector/pg_stat_statements.go
@@ -110,6 +110,18 @@ func NewPGStatStatementsCollector(config collectorConfig) (Collector, error) {
 }
 
 var (
+	statStatementsPlansTotal = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, statStatementsSubsystem, "plans_total"),
+		"Number of times planned",
+		[]string{"user", "datname", "queryid"},
+		prometheus.Labels{},
+	)
+	statStatementsPlansSecondsTotal = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, statStatementsSubsystem, "plans_seconds_total"),
+		"Total time spent in planning the statement, in seconds",
+		[]string{"user", "datname", "queryid"},
+		prometheus.Labels{},
+	)
 	statStatementsCallsTotal = prometheus.NewDesc(
 		prometheus.BuildFQName(namespace, statStatementsSubsystem, "calls_total"),
 		"Number of times executed",
@@ -159,6 +171,8 @@ const (
 		pg_database.datname,
 		pg_stat_statements.queryid,
 		%s
+		0 as plans_total,
+		0 as plans_seconds_total,
 		pg_stat_statements.calls as calls_total,
 		pg_stat_statements.total_time / 1000.0 as seconds_total,
 		pg_stat_statements.rows as rows_total,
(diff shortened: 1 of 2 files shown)


collector/pg_stat_statements.go
```diff
@@ -110,6 +110,18 @@ func NewPGStatStatementsCollector(config collectorConfig) (Collector, error) {
...
+	statStatementsPlansTotal = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, statStatementsSubsystem, "plans_total"),
+		"Number of times planned",
+		[]string{"user", "datname", "queryid"},
+		prometheus.Labels{},
+	)
+	statStatementsPlansSecondsTotal = prometheus.NewDesc(
+		prometheus.BuildFQName(namespace, statStatementsSubsystem, "plans_seconds_total"),
+		"Total time spent in planning the statement, in seconds",
+		[]string{"user", "datname", "queryid"},
```
ArthurSens, line 122: queryID looks like a label with unbounded cardinality. Could we drop it?

jfcoz: I need it, for example to get the ratio of calls by plan and find queries which are badly used (for example prepare + bind inside a loop)

ArthurSens: This doesn't work well with Prometheus though 😅.

Maybe that could become an exemplar label in the histogram?

jfcoz: I don’t understand the problem. It’s already the same in all the other metrics from pg_stat_statements.

ArthurSens: Apologies, I probably should have said this from the beginning. I'm not a maintainer for this repository! If the maintainers are ok with it, that's what counts

ArthurSens: What I meant by "this doesn't work well with Prometheus" is that a single query will have a single ID and generate a single timeseries; another query will get a different ID and therefore become a brand-new timeseries.

All those unique IDs will become an entry in Prometheus's Posting index ([good blog in case you'd like to read more](https://ganeshvernekar.com/blog/prometheus-tsdb-persistent-block-and-its-index#3-index)). A gigantic posting index will make PromQL queries slower and increase resource usage

collector/pg_stat_statements.go
```diff
@@ -159,6 +171,8 @@ const (
 		pg_database.datname,
 		pg_stat_statements.queryid,
 		%s
+		0 as plans_total,
+		0 as plans_seconds_total,
```
ArthurSens, line 175: Not emitting the metric at all sounds better than a 0, people might misinterpret this as the table actually returning 0.

ArthurSens (changes requested):
Hi, getting back to this after a few months of extra experience with this exporter 😅. I'm ok with the queryID now, I understand that this is already present everywhere.

The codebase has been heavily refactored in the last few months (#1330), so a rebase will be needed, or even a fresh new PR, whatever is easier.

#1351 added configuration to preserve correct BIGINT numbers, so we also need to keep this config for the new metrics.


And last question: The collector currently selects its top statements by execution time. A planning-heavy but execution-cheap query (the main use case for this change) could therefore be excluded by the limit. Would ordering by combined planning and execution time, or making the ordering configurable, better serve this feature?

PR #1327, opened 2026-06-05
Fix short_desc null bug with coalesce wrapping

### LINKED ISSUE #1186: Error Scan error on column index 3, name \"short_desc\": converting NULL to string is unsupported
Errors in exporter logs:
ts=2025-08-25T13:38:29.358Z caller=postgres_exporter.go:714 level=error err="error retrieving settings: Error retrieving rows on \"localhost:5432\": pg sql: Scan error on column index 3, name \"short_desc\": converting NULL to string is unsupported"

Postgres Pro Enterprise 16.6 has null values for these parameters:
replan_enable
replan_show_signature

Please fix pg_settings query.

jun-gradial:
Summary:
- Fixes https://github.com/prometheus-community/postgres_exporter/issues/1186.
- `pg_settings.short_desc` can be `NULL` on some PostgreSQL-compatible systems, which causes the collector scan to fail with `converting NULL to string is unsupported` because the value is scanned into a Go string.
- Match the existing nullable `unit` handling by wrapping `short_desc` with `COALESCE(short_desc, '')`, so settings collection can continue when this optional description is missing.

Tests:
- `gofmt -w collector/pg_setting.go collector/pg_setting_test.go`
- `go test ./collector`

### DIFF
--- a/collector/pg_setting.go
+++ b/collector/pg_setting.go
@@ -52,7 +52,7 @@ var (
 	// Settings intentionally ignored due to invalid format:
 	// - `sync_commit_cancel_wait`, specific to Azure Postgres, see https://github.com/prometheus-community/postgres_exporter/issues/523
 	// - `google_dataplex.max_messages`, specific to Google Cloud SQL, see https://github.com/prometheus-community/postgres_exporter/issues/1240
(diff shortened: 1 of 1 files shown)


jun-gradial:
@sysadmind @ArthurSens  could you please take a look at this one ? thx.

ArthurSens:
Looking at the surrounding code, it seems `short_desc` isn't used for anything. Maybe the solution is to remove it from the query entirely 🤔 

I'm not a maintainer here though, let's wait to see what maintainers have to say

sysadmind (approved):
LGTM. Looks like it becomes the HELP for the metric inside *pgSetting.metric()

PR #1330, opened 2026-06-11
Refactor Postgres exporter to make it re-usable as a Go library

ArthurSens:
Superseeds #1297 and #1299

---

This PR refactors the Postgres exporter codebase so it can be easily embedded as a Go library in downstream projects. The architecture I'm aiming for in this PR is documented [here](https://github.com/prometheus/prometheus-opentelemetry-collector/blob/main/docs/embeddable-exporters.md).

To summarize:
* We're abstracting the concept of "One Postgres exporter" into an object called `Runtime`. This `Runtime` contains all Prometheus collectors in isolation.
* We have a `Config` package that holds all possible user-facing configuration, with the exception of `AuthModules`. That includes all CLI flags, including the enabled collectors.
* `cmd/postgres/main.go` stitches CLI flags, config, Runtime, and http handler together
* `cmd/postgres/probes.go` uses the Runtime abstraction to handle multiple postgres targets.

For more details on the architecture, I recommend reading the [docs](https://github.com/prometheus/prometheus-opentelemetry-collector/blob/main/docs/embeddable-exporters.md) we wrote or the Stackdriver exporter codebase, which already follows this architecture!


collector/runtime.go
```diff
@@ -0,0 +1,110 @@
...
+	postgresCollector *collector.PostgresCollector
+}
+
+func NewRuntime(cfg *config.Config, logger *slog.Logger) (*Runtime, error) {
+	if cfg == nil {
+		return nil, errors.New("config is required")
+	}
+	if !cfg.Validated() {
+		return nil, errors.New("config has not been validated; call cfg.Validate before NewRuntime")
+	}
```
kgeckhart: I kinda like what you did here, https://github.com/prometheus/prometheus-opentelemetry-collector/pull/11/changes#diff-e855075d0398417658889b1c58d1abee6934a122028a324919d2549250e16e72R60-R64, to have it call validate. We might want to log a warning since we would hope it would be done ahead of time but avoiding the error seems nice.

collector/runtime.go
sysadmind, line 1: Can we move the runtime.go and runtime_test.go to the collector/ package please? I think it's confusing to have a collectors/ package as a sibling. This has to call into collector/ anyway, so lets put them together.

collector/runtime.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/prometheus-community/postgres_exporter/exporter"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+type Runtime struct {
+	exporter          *exporter.Exporter
+	postgresCollector *collector.PostgresCollector
+}
+
+func NewRuntime(cfg *config.Config, logger *slog.Logger) (*Runtime, error) {
```
sysadmind: Can we change this config to not be a pointer please? We have to validate it, so we don't want anyone changing it.

sysadmind (changes requested):

collector/probe.go
```diff
@@ -31,30 +32,27 @@ type ProbeCollector struct {
 
 func NewProbeCollector(logger *slog.Logger, excludeDatabases []string, registry *prometheus.Registry, dsn config.DSN) (*ProbeCollector, error) {
 	collectors := make(map[string]Collector)
-	initiatedCollectorsMtx.Lock()
-	defer initiatedCollectorsMtx.Unlock()
-	for key, enabled := range collectorState {
+	for key, enabled := range DefaultCollectorStates() {
```
sysadmind: Doesn't this change the behavior so that it's only ever the default collectors that get enabled?

config/config.go
```diff
@@ -20,12 +20,156 @@ import (
...
+
+	c.validated = true
+	return nil
+}
+
+func (c Config) Validated() bool {
+	return c.validated
+}
+
+func DefaultCollectorConfig() map[string]bool {
```
sysadmind, line 175: I don't like this configuration being here. Before, you set all the defaults in the collector/ package. Each collector touches its own file and sets the default and the name. That's it. 

Here , you have to know to add the name here, and to also set the default here? I think that the only place that defaults should be set for collectors is with the collector itself. That package already has a new `DefaultCollectorStates()` that might work, but I don't like the duplication.

config/config.go
```diff
@@ -20,12 +20,156 @@ import (
...
+		if _, ok := DefaultCollectorConfig()[name]; !ok {
+			return fmt.Errorf("unknown collector %q", name)
+		}
+	}
+
+	c.validated = true
+	return nil
+}
+
+func (c Config) Validated() bool {
```
sysadmind: I think we can drop this. We have the above validate() which is all that should matter.

config/config.go
```diff
@@ -20,12 +20,156 @@ import (
...
+const (
+	DefaultMetricPrefix      string        = "pg"
+	DefaultCollectionTimeout time.Duration = time.Minute
+
+	DefaultPGStatStatementsIncludeQuery bool = false
+	DefaultPGStatStatementsQueryLength  uint = 120
+	DefaultPGStatStatementsLimit        uint = 100
+)
+
+const (
```
sysadmind, line 40: I don't like having to define these names here in addition to in collector/

config/config.go
```diff
@@ -20,12 +20,156 @@ import (
...
+	DisableDefaultMetrics bool
+	AutoDiscoverDatabases bool
+	UserQueriesPath       string
+	ConstantLabels        string
+	ExcludeDatabases      []string
+	IncludeDatabases      []string
+	Collectors            map[string]bool
+	PGStatStatements      PGStatStatementsConfig
+
+	validated bool
```
sysadmind: I think we can just remove this bool struct field. Validate() is the only function we should need.

config/config.go
```diff
@@ -20,12 +20,156 @@ import (
...
+}
+
+type PGStatStatementsConfig struct {
```
sysadmind, line 106: What would it look like for this to be built by the collector instead and imported. If we're already importing the package for the defaults, it would mean we can leave config with the collector itself.

collector/probe.go
sysadmind: Actually, I think the reason this wouldn't come up is because this is now dead code. If we're moving away from this to runtime, we should clean up this dead code.

cmd/postgres_exporter/main.go
```diff
@@ -76,7 +131,15 @@ func main() {
 		return
 	}
 
-	if err := c.ReloadConfig(*configFile, logger); err != nil {
+	registry := prometheus.NewRegistry()
+	var err error
+	c, err = config.NewHandler(registry)
```
sysadmind: I find this weird that we pre-define C here. This is an easy thing to miss while reading the code - that C is a package var, not func local. I think this should change to explicitly set the package var, or better yet, can we just not have the package var if we don't need it?

cmd/postgres_exporter/main.go
```diff
@@ -161,11 +212,65 @@ func main() {
...
+	result := make([]string, 0, len(values))
+	for _, value := range values {
+		if trimmed := strings.TrimSpace(value); trimmed != "" {
+			result = append(result, trimmed)
+		}
+	}
+	return result
+}
+
+func mustParseDurationFlag(value string) time.Duration {
```
sysadmind: Func names that start `must` typically panic. In this case, that feels wrong. I think we should drop the `must` prefix.

cmd/postgres_exporter/main.go
```diff
@@ -161,11 +212,65 @@ func main() {
...
+	duration, err := time.ParseDuration(value)
+	if err != nil {
+		// Keep Kingpin's current parse-later behavior but fail clearly during
+		// config validation/construction.
```
sysadmind: This error is silent to the user. I think it should return the error and the calling code should log it.

cmd/postgres_exporter/main.go
sysadmind: Also, does this create an easy dev mistake? We have this as a pointer and the probe handler relies on it too. It leaves potential for a nil pointer.

sysadmind (changes requested):

config/config.go
ArthurSens: Hmmmm, I'm a bit conflicted on this one. I'm aligned with not liking the duplication, but I feel like the default configuration should live in the config package. The OTel Collector's Component interface requires a function that returns default configuration, and having to import the whole Collectors package for this sounds a bit too much 🤔. 

Would you be okay with keeping defaults in the config package, but finding a way to remove duplication?

config/config.go
ArthurSens: Yeah, I'm not 100% here about this. We discussed removing the Config pointer in NewRuntime, but if Config has pointer fields, like maps, that means even the value copy is subject to mutations.

It seems like we need to do getters and setters after all...? :(

config/config.go
ArthurSens: Yeah me neither, I was trying to avoid unnecessary package imports, but it might be better than duplication 🤷

config/config.go
ArthurSens: You mean we should always call Validate() inside NewRuntime(), rather than outside?

cmd/postgres_exporter/main.go
ArthurSens: removed the package global :)

config/config.go
```diff
@@ -18,14 +18,189 @@ import (
...
+	}
+	if c.CollectionTimeout <= 0 {
+		return ValidatedConfig{}, fmt.Errorf("collection timeout must be greater than zero")
+	}
+	for i, dsn := range c.DataSourceNames {
+		if dsn == "" {
+			return ValidatedConfig{}, fmt.Errorf("data source name at index %d must not be empty", i)
+		}
+	}
+	if c.PGStatStatements.QueryLength == 0 {
```
sysadmind: The error message says greater than - I think this test should also be greater than right?

config/config.go
```diff
@@ -18,14 +18,189 @@ import (
...
+		return ValidatedConfig{}, fmt.Errorf("pg_stat_statements query length must be greater than zero")
+	}
+	if c.PGStatStatements.Limit == 0 {
```
sysadmind: Here too on greater than.

config/config.go
```diff
@@ -18,14 +18,189 @@ import (
...
+	for name := range c.Collectors {
+		if name == "" {
+			return ValidatedConfig{}, fmt.Errorf("collector name must not be empty")
+		}
+		if _, ok := DefaultCollectorConfig()[name]; !ok {
+			return ValidatedConfig{}, fmt.Errorf("unknown collector %q", name)
+		}
+	}
+
+	return ValidatedConfig{inner: c.clone(), ok: true}, nil
```
sysadmind: Maybe this is a bit too careful, but what if we call c.clone() at the top of this function? That means the checks are all against our own personal copy and there's no risk of anyone modifying it. I don't think it really costs much extra. We're going to do all the checking. The only extra overhead is the copy if we're going to error.

ArthurSens: I think that's a good idea, if we clone at the end then we're still at risk of mutations

ArthurSens:
@sysadmind, thanks a lot for all the reviews so far! I believe all comments are addressed

sysadmind (approved):
I think this works the way it is. It does mean that contributors need to know more and touch more files to add a new collector. Maybe we should document that in the contributing guide.

PR #1351, opened 2026-07-24
fix: preserve precision for large counters

### LINKED ISSUE #300: Conversion from int64 to float64 prometheus metrics not ideal for counters
Currently all int64 metrics are converted to prometheus metrics using:
```
	case int64:
		return float64(v), true
```

That's unfortunate because it loses precision, any value over 2^53 will be imprecise.

For gauges this is still the best we can do, and the loss of precision may not matter much anyways. But for counters we can do better since resets are handled. And moreover for counters the loss of precision could be a major issue since it's the difference between two values which matters.

I would suggest we convert bigints to float64 for prometheus by just taking them mod 2^53. That means we effectively always use the full precision of the mantissa and never lose precision. For values larger than 2^53 the counter will just wrap around to 0 and but that's ok for counters, rate() et al will still work correctly.

So if you just made that something like `return float64(v % uint64(1<<53))`, at least for counters, then there would be no loss in precision for larger values.

gnanirahulnutakki:
## Summary

- wrap non-negative `int64` counter samples modulo 2^53 before converting them to `float64`
- apply the conversion to both legacy metric maps and the refactored `collector` runtime
- keep gauges, negative values, and other metric types unchanged
- enable wrapping by default, with `--no-wrap-large-counters` as a CLI-only rollout opt-out
- document the upgrade reset boundary prominently in the changelog

## Why

Prometheus samples are `float64`, so adjacent integer values above 2^53 can collapse to the same value during conversion. For counters, wrapping at 2^53 preserves single-unit deltas. This follows the existing approach in `prometheus/snmp_exporter`.

Fixes #300.

## Compatibility warning

Enabling wrapping changes the first sample of a counter already above 2^53. For example, `2^53 + 1,234,567` becomes `1,234,567`; Prometheus interprets that decrease as a reset and counts the remainder as post-reset growth, which can create a one-time false `rate()`/`increase()` spike during upgrade. The changelog now calls this out, and `--no-wrap-large-counters` preserves the previous conversion during a staged rollout.

## Validation

- rebased onto upstream `master` at `90a293c`
- `make GO_ONLY=1 SKIP_GOLANGCI_LINT=1` in the project's `quay.io/prometheus/golang-builder:1.26-base` image (Go 1.26.7), including the full race suite
- `golangci-lint` v2.12.2: zero issues
- `go test -race -tags integration ./... -count=1` against PostgreSQL 13.22, 14.19, 15.14, 16.10, 17.6, and 18.1
- live PostgreSQL 18.1 scrape: default exports the warning example as `1.234567e+06`; `--no-wrap-large-counters` exports the legacy `9.00719925597556e+15`; both report `pg_up 1` with zero scrape/query-load errors

### DIFF
--- a/cmd/postgres_exporter/main.go
+++ b/cmd/postgres_exporter/main.go
@@ -49,6 +49,7 @@ var (
 	includeDatabases      = kingpin.Flag("include-databases", "A list of databases to include when autoDiscoverDatabases is enabled (DEPRECATED)").Default("").Envar("PG_EXPORTER_INCLUDE_DATABASES").String()
 	metricPrefix          = kingpin.Flag("metric-prefix", "A metric prefix can be used to have non-default (not \"pg\") prefixes for each of the metrics").Default("pg").Envar("PG_EXPORTER_METRIC_PREFIX").String()
 	collectionTimeout     = kingpin.Flag("collection-timeout", "Timeout for collecting the statistics when the database is slow").Default("1m").Envar("PG_EXPORTER_COLLECTION_TIMEOUT").String()
+	wrapLargeCounters     = kingpin.Flag("wrap-large-counters", "Wrap 64-bit counters at 2^53 to avoid floating point rounding.").Default("true").Envar("PG_EXPORTER_WRAP_LARGE_COUNTERS").Bool()
 	logger                = promslog.NewNopLogger()
 )
 
@@ -110,6 +111,7 @@ func main() {
 		exporter.ExcludeDatabases(excludedDatabases),
 		exporter.IncludeDatabases(*includeDatabases),
 		exporter.WithMetricPrefix(*metricPrefix),
+		exporter.WrapLargeCounters(*wrapLargeCounters),
 	}
 
 	exporter := exporter.NewExporter(dsns, logger, opts...)
--- a/cmd/postgres_exporter/main.go
+++ b/cmd/postgres_exporter/main.go
@@ -51,6 +51,7 @@ var (
 	includeDatabases      = kingpin.Flag("include-databases", "A list of databases to include when autoDiscoverDatabases is enabled (DEPRECATED)").Default("").Envar("PG_EXPORTER_INCLUDE_DATABASES").String()
 	metricPrefix          = kingpin.Flag("metric-prefix", "A metric prefix can be used to have non-default (not \"pg\") prefixes for each of the metrics").Default("pg").Envar("PG_EXPORTER_METRIC_PREFIX").String()
 	collectionTimeout     = kingpin.Flag("collection-timeout", "Timeout for collecting the statistics when the database is slow").Default("1m").Envar("PG_EXPORTER_COLLECTION_TIMEOUT").String()
+	wrapLargeCounters     = kingpin.Flag("wrap-large-counters", "Wrap 64-bit counters at 2^53 to avoid floating point rounding.").Default("true").Envar("PG_EXPORTER_WRAP_LARGE_COUNTERS").Bool()
 	collectorFlags        = newCollectorFlags()
 	statStatementsFlags   = newPGStatStatementsFlags()
 	logger                = promslog.NewNopLogger()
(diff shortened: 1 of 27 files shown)


cmd/postgres_exporter/main.go
```diff
@@ -49,6 +49,7 @@ var (
 	includeDatabases      = kingpin.Flag("include-databases", "A list of databases to include when autoDiscoverDatabases is enabled (DEPRECATED)").Default("").Envar("PG_EXPORTER_INCLUDE_DATABASES").String()
 	metricPrefix          = kingpin.Flag("metric-prefix", "A metric prefix can be used to have non-default (not \"pg\") prefixes for each of the metrics").Default("pg").Envar("PG_EXPORTER_METRIC_PREFIX").String()
 	collectionTimeout     = kingpin.Flag("collection-timeout", "Timeout for collecting the statistics when the database is slow").Default("1m").Envar("PG_EXPORTER_COLLECTION_TIMEOUT").String()
+	wrapLargeCounters     = kingpin.Flag("wrap-large-counters", "Wrap 64-bit counters at 2^53 to avoid floating point rounding.").Default("true").Envar("PG_EXPORTER_WRAP_LARGE_COUNTERS").Bool()
```
SuperQ: We generally avoid Env vars in the Prometheus community. This exporter has a bunch for legacy reasons. Let's not continue the trend.


```suggestion
	wrapLargeCounters     = kingpin.Flag("wrap-large-counters", "Wrap 64-bit counters at 2^53 to avoid floating point rounding.").Default("true").Bool()
```

SuperQ (dismissed):

cmd/postgres_exporter/main.go
gnanirahulnutakki: Addressed in 9c8b1d9: removed the new `PG_EXPORTER_WRAP_LARGE_COUNTERS` environment-variable binding and its README entry while retaining the default-on CLI flag and `--no-wrap-large-counters`. Validation passed with `go test ./... -count=1`, `make unused`, `make build test`, and golangci-lint (0 issues).

ArthurSens (commented):
Thanks! However the fix is a bit more involving than what you've worked so far.

You fixed the problem on the metrics that are collected in the "legacy" way, which we're slowly migrating away from. Today, most of the metric collection happens in the `collector` package. Over there, you'll see several metrics, [like this one](https://github.com/prometheus-community/postgres_exporter/blob/5b801d2bfc4c05d2867d5495b43b630057327751/collector/pg_stat_bgwriter.go#L132-L143), are still doing an Int64 to Float64 conversion.

gnanirahulnutakki:
Thanks for pointing out the collector path. I extended the fix in 52cae0f to cover the Int64-based CounterValue conversions in the collector package and threaded --no-wrap-large-counters through both the normal and /probe collector paths. I intentionally left gauges and Float64 duration/time metrics unchanged.\n\nI validated exact head 52cae0f with Go 1.26:\n\n- go test -race ./... -count=1\n- CI-equivalent build/test, golangci-lint v2.12.2, and yamllint\n- uncached integration suite against PostgreSQL 13.22, 14.19, 15.14, 16.10, 17.6, and 18.1\n- live /metrics scrapes on 13.22 and 18.1, asserting pg_up == 1 and pg_stat_database_xact_commit is emitted\n\nCould you take another look at the collector coverage?

ArthurSens:
Hey @gnanirahulnutakki, sorry for the delay and trouble here. Looks like postgres exporter went through a refactoring and this PR will need to go through a rebase.

I also think this change deserves a big warning in the changelog because enabling it by default can cause a large false-rate spike during upgrade. If the previous sample is `2^53 + 1,234,567`, the first wrapped sample is `1,234,567`. Prometheus interprets the decrease as a reset and counts that remainder as post-reset growth—not merely a harmless reset.

gnanirahulnutakki:
Addressed on `61cc12f`:

- rebased onto current `master` (`90a293c`) and adapted the flag through the new `config.ValidatedConfig` / `collector.Runtime` path, including normal and `/probe` collection;
- added the prominent changelog warning with the exact `2^53 + 1,234,567` reset/spike example;
- added that same value as a regression case rather than leaving the warning as prose only.

Validation on the rebased head:

- project CI-equivalent `make GO_ONLY=1 SKIP_GOLANGCI_LINT=1` in `quay.io/prometheus/golang-builder:1.26-base` (Go 1.26.7), including the full race suite: pass;
- `golangci-lint` v2.12.2: 0 issues;
- race-enabled integration suite: pass on PostgreSQL 13.22, 14.19, 15.14, 16.10, 17.6, and 18.1;
- live PostgreSQL 18.1 scrape: default produced `large_counter_proof_value 1.234567e+06`, while `--no-wrap-large-counters` produced `9.00719925597556e+15`; both had `pg_up 1` and zero scrape/query-load errors.

ArthurSens (approved):
LGTM, thank you!

PR #1353, opened 2026-07-25
stat_replication: add pid label to disambiguate replication connections

### LINKED ISSUE #1352: [0.20.1] 2 metrics with "was collected before with the same name and label values" error
**What did you do?**
I upgraded postgres exxporter from 0.20.0 to 0.20.1 /metrics is returning a http 500 error

http 200 Ok with a list of metrics

```
An error has occurred while serving metrics:

2 error(s) occurred:
* collected metric "pg_stat_replication_pg_current_wal_lsn_bytes" { label:{name:"application_name" value:"walreceiver"} label:{name:"client_addr" value:"172.18.0.254/32"} label:{name:"slot_name" value:""} label:{name:"state" value:"streaming"} gauge:{value:4.5350988e+08}} was collected before with the same name and label values
* collected metric "pg_stat_replication_pg_wal_lsn_diff" { label:{name:"application_name" value:"walreceiver"} label:{name:"client_addr" value:"172.18.0.254/32"} label:{name:"slot_name" value:""} label:{name:"state" value:"streaming"} gauge:{value:0}} was collected before with the same name and label values
```

**Environment**

* System information:
**Host**: Synology DS1522+ DSM 7.3.2-86009 Update 4  Linux 4.4.302+ x86_64
**Docker**: Docker version 24.0.2, build 610b8d0

* postgres_exporter version:
Docker Container
postgres_exporter, version 0.20.1 (branch: HEAD, revision: 867fbcac31cd18c143e244190ea9168cca069827)
  build user:       root@f281ca536a03
  build date:       20260708-01:46:52
  go version:       go1.26.4
  platform:         linux/amd64
  tags:             unknown
* postgres_exporter flags:

```
no flag, default configuration using env variables to define the URI
```

* PostgreSQL version:
Docker Container, same stack as postgres_exporter, same docker network
psql (PostgreSQL) 18.4 (Debian 18.4-1.pgdg13+1)

Postgres DB is the master and have replication enabled

* Logs:
```
No logs in postgres or postgres_exporter linked to this bug
```

Reverting from 0.20.1 to 0.20.0 solves the issue without any other modification

sysadmind:
application_name, client_addr, state, and slot_name are not a unique key for pg_stat_replication rows. Multiple standbys with the same application_name, client_addr, state, and slot_name (empty) were causing duplicate metrics which would get rejected by the registry.

This regressed visibly in 0.20.1: 0.20.0 had the stat_replication query erroring on every scrape ("column slot_name does not exist", #1310), so the collector never actually emitted rows. #1313 fixed that query error, which let real data flow again and exposed this pre-existing label collision for the first time.

One downside is that the additional label could increase series churn because every time the walsender reconnects, the pid will change and therefore create a new series.

Fixes #1352

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## main / (unreleased)
 
-* [CHANGE] stat_replication: add `pid` label to disambiguate replication connections that otherwise share identical labels by @sysadmind in https://github.com/prometheus-community/postgres_exporter/pull/TODO
+* [CHANGE] stat_replication: add `pid` label to disambiguate replication connections that otherwise share identical labels by @sysadmind in https://github.com/prometheus-community/postgres_exporter/pull/1353
 
 ## 0.20.1 / 2026-07-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## main / (unreleased)
 
+* [CHANGE] stat_replication: add `pid` label to disambiguate replication connections that otherwise share identical labels by @sysadmind in https://github.com/prometheus-community/postgres_exporter/pull/TODO
+
 ## 0.20.1 / 2026-07-07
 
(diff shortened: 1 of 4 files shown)


SuperQ (approved):
LGTM

PR #1361, opened 2026-08-08
Create user facing documentation

sysadmind:
Create a docs/ directory with user facing documentation. This covers getting started, running with docker, specifying secrets, configuration, and database permissions. This is all focused on the perspective of a user or someone deploying the exporter.


nicolastakashi:
Is part of #1363

ArthurSens (commented):
I haven't read the written documentation just yet, I'm just thinking about how this would be integrated in the website.

Since you're writing separate documents per topic, are you imagining that each document also becomes a separate page in the website? With README kinda working as some kind of index.html for postgres exporter?

docs/docker.md
```diff
@@ -0,0 +1,59 @@
...
+- Quay.io: [`quay.io/prometheuscommunity/postgres-exporter`](https://quay.io/repository/prometheuscommunity/postgres-exporter)
+- GHCR: [`ghcr.io/prometheus-community/postgres-exporter`](https://github.com/prometheus-community/postgres_exporter/pkgs/container/postgres-exporter)
+
+All three are the same image; there's no functional difference between them — use whichever registry your infrastructure prefers. Note that the GHCR organization is `prometheus-community` (hyphenated, matching the GitHub org), while Docker Hub and Quay use `prometheuscommunity`.
+
+## Tags
+
+- `latest` — the most recent release.
+- `vX.Y.Z` — a specific release, e.g. `v0.18.1`. Pin to a specific tag for production deployments so upgrades are intentional.
+- `master` — the latest build from the `master` branch, which may be unstable. Use this only for testing or development.
```
ArthurSens, line 15: We don't need to explain semver here 🤔

docs/docker.md
```diff
@@ -0,0 +1,59 @@
+# Docker Images
```
ArthurSens, line 1: I wouldn't create a document called `Docker`, especially when Docker is already mentioned in the Getting Started.

Maybe in Getting Started, we could have different sections? Getting Started with Docker/Kubernetes/Binary/Ansible/whatever is available.

I'm imagining a component like this (change Java/Go with what is relevant to us)

Image

Where the reader can click and choose the preferred way of getting started

docs/configuration.md
```diff
@@ -0,0 +1,82 @@
...
+| `stat_progress_vacuum` | enabled | |
+| `stat_replication` | enabled | |
+| `stat_statements` | disabled | requires the `pg_stat_statements` extension |
+| `stat_user_tables` | enabled | |
+| `stat_wal_receiver` | disabled | |
+| `statio_user_indexes` | disabled | |
+| `statio_user_tables` | enabled | |
+| `wal` | enabled | |
+| `xlog_location` | disabled | |
+| `buffercache_summary` | disabled | requires the `pg_buffercache` extension |
```
ArthurSens: I'd love it if this table could be automated with [mdox](https://github.com/bwplotka/mdox). We should have a hidden CLI flag that builds this table to make sure it is always up to date.

docs/configuration.md
```diff
@@ -0,0 +1,82 @@
...
+auth_modules:
+  foo: # referenced as ?auth_module=foo on /probe requests
+    type: userpass
+    userpass:
+      username: monitoring_user
+      password: monitoring_password
+    options:
+      # merged into the DSN as key=value query parameters
+      sslmode: require
+```
```
ArthurSens, line 127: This one also, it would be awesome if we used `mdox-exec=cat example.yaml` and we actually use this file as part of our tests in CI to make sure it's always valid.

docs/connecting.md
```diff
@@ -0,0 +1,60 @@
+# Connecting to PostgreSQL
```
ArthurSens, line 1: Isn't database permissions related to this topic? should we merge them?

docs/getting-started.md
```diff
@@ -0,0 +1,43 @@
...
+
+```bash
+curl "http://localhost:9187/metrics"
+```
+
+You should see PostgreSQL metrics prefixed with `pg_`, along with the exporter's own internal metrics (`go_*`, `promhttp_*`, etc).
+
+## Point Prometheus at it
+
+Add a scrape job to your Prometheus configuration:
```
nicolastakashi, line 80: Maybe worth a cross reference here pointing to the scrape config docs for more details, mentioning this is a simple example, wdyt?

docs/configuration.md
nicolastakashi: I also believe that we should have some meaninfull description for each option

docs/configuration.md
```diff
@@ -0,0 +1,82 @@
...
+| `statio_user_tables` | enabled | |
+| `wal` | enabled | |
+| `xlog_location` | disabled | |
+| `buffercache_summary` | disabled | requires the `pg_buffercache` extension |
+
+`pg_stat_statements` has its own sub-flags: `--collector.stat_statements.include_query` (off by default — includes raw query text), `--collector.stat_statements.query_length` (default 120), `--collector.stat_statements.limit` (default 100), and `--collector.stat_statements.exclude_databases` / `--collector.stat_statements.exclude_users` (comma-separated).
+
+Run `postgres_exporter --help` for the exact, current list of flags — new collectors are added over time and this table can drift.
+
+## Common flags
```
nicolastakashi, line 83: Are all those common flags comming from exporter toolkit? should we reference it here?

docs/configuration.md
```diff
@@ -0,0 +1,82 @@
...
+| `stat_statements` | disabled | requires the `pg_stat_statements` extension |
+| `stat_user_tables` | enabled | |
+| `stat_wal_receiver` | disabled | |
+| `statio_user_indexes` | disabled | |
+| `statio_user_tables` | enabled | |
+| `wal` | enabled | |
+| `xlog_location` | disabled | |
+| `buffercache_summary` | disabled | requires the `pg_buffercache` extension |
+
+`pg_stat_statements` has its own sub-flags: `--collector.stat_statements.include_query` (off by default — includes raw query text), `--collector.stat_statements.query_length` (default 120), `--collector.stat_statements.limit` (default 100), and `--collector.stat_statements.exclude_databases` / `--collector.stat_statements.exclude_users` (comma-separated).
```
nicolastakashi: Should we have a dedicated section to what's pg_stat_statements?

docs/configuration.md
```diff
@@ -0,0 +1,82 @@
...
+    options:
+      # merged into the DSN as key=value query parameters
+      sslmode: require
+```
+
+The config file is reloaded without restarting the process by sending an HTTP POST to `/-/reload`.
+
+## Deprecated options
+
+These remain for backward compatibility but shouldn't be used in new deployments:
```
nicolastakashi, line 133: Aren't those removed on newer versions?

docs/connecting.md
```diff
@@ -0,0 +1,60 @@
...
+Send the target as a query parameter:
+
+```
+http://exporter:9187/probe?target=my-db-host:5432
+```
+
+Prometheus is typically configured to do this via relabeling so the target list can come from service discovery:
+
+```yaml
+scrape_configs:
```
nicolastakashi, line 34: same: https://github.com/prometheus-community/postgres_exporter/pull/1361/changes#r3778534453

docs/configuration.md
ArthurSens: could we have that abstraction in exporter-toolkit @nicolastakashi? :) 

Not sure how yet, but I'm thking something about exporter toolkit managing collector enable/disable states, we could use that to have exporter toolkit automatically generate tables like this for documentation and us using the same state to enable/disable collectors.

Either way, this is work for the future :p

docs/docker.md
sysadmind: My intent here was to list the tags that a normal user might care about. Discoverability of tags for docker images isn't always straightforward - you often have to go dig through a git repo or just read through Dockerhub tags to find what you're looking for. 

Also, while our releases are formatted like semver (x.y.z), we don't really treat it as semver. We will make breaking changes on a Y update. And if you think of it like a library (which you will going forward) the PR you just had to make it embeddable is probably a breaking change even though it was not user visible.

sysadmind:
@ArthurSens  yes I'm imagining each file becomes a navigation item in the site. One thing I also want to consider with this - people will also come directly to this repo to view docs, so I want to enable viewing them in Github and avoid anything too custom that reduces the value of them as standalone files in this repo.

docs/docker.md
sysadmind: I think that there's value here. The docker file talks about a lot more than you need just to get started. The image details are something that a user would have to go digging through code to understand if it's not called out.

docs/configuration.md
sysadmind: Many are or from prometheus/common. Updated this section.

ArthurSens:
I spent some time looking at how to pull exporter docs into prometheus.io. The docs site can support this with a latest-release-tag source for exporters, but each Markdown file needs the same frontmatter contract that the site uses for title/nav ordering.

Could you add this frontmatter while keeping the existing `#` heading in each file? Keeping the H1 means the files still read well on GitHub; the website will use the frontmatter for metadata and navigation.

`docs/README.md`:

```yaml
---
title: postgres_exporter Documentation
nav_title: Overview
sort_rank: 1
---
```

`docs/getting-started.md`:

```yaml
---
title: Getting Started
sort_rank: 2
---
```

`docs/connecting.md`:

```yaml
---
title: Connecting to PostgreSQL
sort_rank: 3
---
```

`docs/database-permissions.md`:

```yaml
---
title: Database Permissions
sort_rank: 4
---
```

`docs/configuration.md`:

```yaml
---
title: Configuring the Exporter
nav_title: Configuration
sort_rank: 5
---
```

`docs/secrets.md`:

```yaml
---
title: Secrets
sort_rank: 6
---
```

`docs/docker.md`:

```yaml
---
title: Docker Images
nav_title: Docker
sort_rank: 7
---
```

`docs/aws-rds.md`:

```yaml
---
title: Running Against AWS RDS
nav_title: AWS RDS
sort_rank: 8
---
```

One important detail: prometheus.io will pull exporter docs from the latest stable release tag, not directly from `master`. So after this merges, the docs will appear on the website once postgres_exporter cuts a release containing the `docs/` directory.

sysadmind:
@ArthurSens I have added the frontmatter. I think this is good for a v1 now.

docs/configuration.md
```diff
@@ -0,0 +1,143 @@
...
+    type: userpass
+    userpass:
+      username: monitoring_user
+      password: monitoring_password
+    options:
+      # merged into the DSN as key=value query parameters
+      sslmode: require
+```
+
+The config file is reloaded without restarting the process by sending an HTTP POST to `/-/reload`.
```
ArthurSens, line 134: Looking at the code, I don't see any `/-/reload` endpoint 🤔. Sounds like a good feature, honestly.

Maybe we should implement it before merging this docs PR?

docs/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+- **Connections and activity** — backend counts by state, long-running transactions, idle sessions, autovacuum workers.
+- **Database-level statistics** — commits, rollbacks, block reads/hits, deadlocks, temp file usage, size on disk.
+- **Table and index I/O** — sequential vs. index scans, tuple counts, dead tuples, vacuum/analyze times.
+- **Replication** — replica lag, replication slots, WAL receiver state, WAL generation.
+- **Checkpointing and background writer** — checkpoint counts and timings, buffers written.
+- **Query statistics** — per-query call counts and timings via the `pg_stat_statements` extension (opt-in).
+- **Server configuration** — the values of `pg_settings`, exported as metrics so configuration drift is queryable.
+
+All metrics are prefixed with `pg_` by default (configurable with `--metric-prefix`). The exporter's own health is exposed via `pg_up` and `pg_exporter_*` metrics alongside the standard Go runtime metrics.
```
ArthurSens, line 24:
```suggestion
All metrics are prefixed with `pg_`. The exporter's own health is exposed via `pg_up` and `pg_exporter_*` metrics alongside the standard Go runtime metrics.
```

--metrix-prefix only changes legacy metrics, not the new ones from the `collectors/` package.

docs/secrets.md
```diff
@@ -0,0 +1,59 @@
...
+```yaml
+auth_modules:
+  prod:
+    type: userpass
+    userpass:
+      username: monitoring_user
+      password: monitoring_password
+```
+
+This keeps credentials out of Prometheus's target list and scrape URLs (which otherwise show up in Prometheus's UI and logs). See [Connecting to PostgreSQL](connecting.md#multi-target-mode-probe-beta) for the full request flow. Note that the config file itself contains plaintext passwords, so its file permissions and any secrets-management layer around it (e.g. templating it from Vault) matter just as much as protecting `DATA_SOURCE_PASS_FILE`.
```
ArthurSens, line 45: `connecting.md#multi-target-mode-probe-beta` looks like a broken link

docs/configuration.md
```diff
@@ -0,0 +1,143 @@
...
+## Common flags
+
+Flags specific to this exporter:
+
+| Flag | Purpose |
+|---|---|
+| `--web.telemetry-path` | Metrics path. Default `/metrics`. |
+| `--config.file` | Path to the [config file](#config-file). Default `postgres_exporter.yml`. |
+| `--collection-timeout` | Per-scrape timeout. Default `1m`. See [Connecting](connecting.md#connection-timeout). |
+| `--metric-prefix` | Prefix for emitted metrics. Default `pg`. |
```
ArthurSens, line 97: #1351 and #1379 introduced new flags that we may want to include here.

I'm also not sure if we want to document the legacy `--metric-prefix` flag 🤔

docs/configuration.md
sysadmind: We can follow up to create this. I pulled the line out of documentation for now. I don't want to confuse people.

sysadmind:
@ArthurSens I cleaned up the most recent review comments with a couple new commits. Can you take a quick once over? We can merge it, but without an approval, the mergability rules aren't met.

docs/configuration.md
```diff
@@ -0,0 +1,143 @@
...
+## Common flags
+
+Flags specific to this exporter:
+
+| Flag | Purpose |
+|---|---|
+| `--web.telemetry-path` | Metrics path. Default `/metrics`. |
+| `--config.file` | Path to the [config file](#config-file). Default `postgres_exporter.yml`. |
+| `--collection-timeout` | Per-scrape timeout. Default `1m`. See [Connecting](connecting.md#connection-timeout). |
+| `--wrap-large-counters` | Wrap 64-bit counters at 2^53 to avoid floating point rounding. Default `false`. |
```
ArthurSens: Default is true
https://github.com/prometheus-community/postgres_exporter/blob/e7e2095249dc369d943af1cef3d8c615228278ec/cmd/postgres_exporter/main.go#L54

docs/aws-rds.md
```diff
@@ -0,0 +1,29 @@
...
+
+Follow the standard non-superuser [database permissions](database-permissions.md) setup, granting `pg_monitor` to your exporter role. Since you don't have a true superuser on RDS, run the grant as your RDS master user, and grant the exporter role to the master user first if needed (see the note in [database-permissions.md](database-permissions.md#postgresql-10-use-the-built-in-pg_monitor-role)).
+
+## `pg_stat_statements`
+
+To use the `stat_statements` collector, `pg_stat_statements` must be preloaded via the RDS parameter group, not just created as an extension:
+
+1. In the RDS parameter group attached to your instance, set:
+   ```
+   shared_preload_libraries = "pg_stat_statements,pg_hint_plan"
```
ArthurSens: is `pg_hint_plan` really needed for RDS?

ArthurSens (approved):
Found a couple extra weird things, but approving! Feel free to address before or after the merge :)

PR #1368, opened 2026-08-18
main: start the exporter through the toolkit bootstrap package

nicolastakashi:
## Summary

Starts `postgres_exporter` on the shared exporter startup path by replacing the hand-rolled `main()` block with `exporter-toolkit/bootstrap`, mirroring prometheus/node_exporter#3660.

The entrypoint currently reimplements what `bootstrap` provides: the `--web.telemetry-path` flag, `kingpinflag`/promslog wiring, version output, landing page construction, and the `web.ListenAndServe` call. Keeping a private copy is how exporters drift apart in flags and startup behavior.

Two notes for review:

- `--dumpmaps` exits from inside the handler factory, since it is an early-exit path and `bootstrap` has no pre-serve hook
- `CloseServers()` now runs after `Run()` returns instead of via `defer`; the old `defer` never actually ran, because every exit path went through `os.Exit`

### DIFF
--- a/cmd/postgres_exporter/main.go
+++ b/cmd/postgres_exporter/main.go
@@ -14,6 +14,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	_ "net/http/pprof"
@@ -32,16 +33,12 @@ import (
 	versioncollector "github.com/prometheus/client_golang/prometheus/collectors/version"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/promslog"
-	"github.com/prometheus/common/promslog/flag"
-	"github.com/prometheus/common/version"
+	"github.com/prometheus/exporter-toolkit/bootstrap"
 	"github.com/prometheus/exporter-toolkit/web"
-	"github.com/prometheus/exporter-toolkit/web/kingpinflag"
 )
 
 var (
 	configFile            = kingpin.Flag("config.file", "Postgres exporter configuration file.").Default("postgres_exporter.yml").String()
-	webConfig             = kingpinflag.AddFlags(kingpin.CommandLine, ":9187")
-	metricsPath           = kingpin.Flag("web.telemetry-path", "Path under which to expose metrics.").Default("/metrics").Envar("PG_EXPORTER_WEB_TELEMETRY_PATH").String()
 	disableDefaultMetrics = kingpin.Flag("disable-default-metrics", "Do not include default metrics.").Default("false").Envar("PG_EXPORTER_DISABLE_DEFAULT_METRICS").Bool()
 	autoDiscoverDatabases = kingpin.Flag("auto-discover-databases", "Whether to discover the databases on a server dynamically. (DEPRECATED)").Default("false").Envar("PG_EXPORTER_AUTO_DISCOVER_DATABASES").Bool()
 	queriesPath           = kingpin.Flag("extend.query-path", "Path to custom queries to run. (DEPRECATED)").Default("").Envar("PG_EXPORTER_EXTEND_QUERY_PATH").String()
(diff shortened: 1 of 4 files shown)


cmd/postgres_exporter/main.go
```diff
@@ -64,16 +61,41 @@ func newConfigHandler() *config.Handler {
...
-	flag.AddFlags(kingpin.CommandLine, promslogConfig)
-	kingpin.HelpFlag.Short('h')
-	kingpin.Parse()
-	logger = promslog.New(promslogConfig)
+	runner := bootstrap.New(bootstrap.Config{
+		App:              kingpin.CommandLine,
+		Name:             exporterName,
+		Description:      "Prometheus PostgreSQL server Exporter",
+		DefaultAddress:   ":9187",
+		MetricsPathEnvar: "PG_EXPORTER_WEB_TELEMETRY_PATH",
```
ArthurSens: I believe we want to deprecate the usage of env vars to configure collectors. Is that right @SuperQ @sysadmind ?

If we already want to deprecate this, it's probably better to not introduce this at all 🤔

sysadmind: Yeah for something like this, I'm in favor of dropping environment variables. I think there are some good use cases for them, but not this one.

cmd/postgres_exporter/main.go
```diff
@@ -175,59 +212,41 @@ func main() {
...
-	)
-	for _, collector := range pgRuntime.Collectors() {
+	pgRuntime = runtime
+
+	if !b.DisableExporterMetrics {
+		registry.MustRegister(
+			prometheuscollectors.NewGoCollector(),
+			prometheuscollectors.NewProcessCollector(prometheuscollectors.ProcessCollectorOpts{}),
+		)
+	}
```
ArthurSens, line 229: Hmmm, I'm just seeing this now. It looks weird that Bootstrap owns the self-metrics toggle but has absolutely no control over it. I wonder if this even is something that the exporter toolkit can control 🤔 

If we compare what this boolean does here with what it does in node exporter, in node exporter we also have HTTP metrics since it uses promhttp.InstrumentMetricsHander 

Version collector is also about the exporter itself and is not gated by the boolean

cmd/postgres_exporter/main.go
```diff
@@ -175,59 +212,41 @@ func main() {
...
-			},
-		}
-		landingPage, err := web.NewLandingPage(landingConfig)
-		if err != nil {
-			logger.Error("error creating landing page", "err", err)
-			os.Exit(1)
-		}
-		http.Handle("/", landingPage)
-	}
+	b.HandleFunc("/probe", handleProbe(logger, authHandler, cfg))
```
ArthurSens, line 235: --web.max-requests works for /metrics but doesn't wrap /probe

cmd/postgres_exporter/main.go
```diff
@@ -175,59 +212,41 @@ func main() {
...
+	// net/http/pprof and the transitively-imported "expvar" package both
+	// register handlers under /debug/ on the default mux.
```
ArthurSens, line 237: this comment seems unnecessary

ArthurSens (commented):
I think this is also a great opportunity to add a big warning in the changelog about the deprecation of env vars!

cmd/postgres_exporter/main.go
nicolastakashi: Yup good catch, I moved version collector under the flag for now. Longer term I think this belongs in bootstrap itself, it owns the registry and gates self-metrics directly instead of just handing exporters a bool, happy to open that as a follow-up on exporter-toolkit if you're up for it wdyt?

cmd/postgres_exporter/main.go
nicolastakashi: Yeah, makes sense, maybe we should have exporter toolkit enforce web.max-requests on every route, wdyt? if you agree I'd make it on the exporter instead of rebuilding it here.

ArthurSens: I think for now we can just do it here. I'm not sure if it's safe to do that, we can make this move later :)

cmd/postgres_exporter/main.go
ArthurSens: I don't know if that's possible though. If it is, then great...let's do it

Issue #1374, opened 2026-08-22
Document --datasource.* flags and mark env vars deprecated


nicolastakashi:
Part of #1371. Depends on #1372, #1373.

**Summary**

Update documentation so the new CLI flags are the documented way to configure `postgres_exporter`, and the env vars are clearly marked deprecated with a migration path.

Issue #1373, opened 2026-08-22
Add deprecation warnings for DATA_SOURCE_* and PG_EXPORTER_* env vars


nicolastakashi:
Part of #1371. Depends on #1372.

**Summary**

Once flag equivalents exist for all env vars (see #1372 for the `DATA_SOURCE_*` datasource flags; the `PG_EXPORTER_*` flags already have equivalents via kingpin's `Envar()`), log a warning at startup whenever an env var is used instead of its flag, pointing to the flag that should replace it.

**Env vars to cover**

- `Envar()`-backed: `PG_EXPORTER_WEB_TELEMETRY_PATH`, `PG_EXPORTER_DISABLE_DEFAULT_METRICS`, `PG_EXPORTER_AUTO_DISCOVER_DATABASES`, `PG_EXPORTER_EXTEND_QUERY_PATH`, `PG_EXPORTER_CONSTANT_LABELS`, `PG_EXPORTER_EXCLUDE_DATABASES`, `PG_EXPORTER_INCLUDE_DATABASES`, `PG_EXPORTER_METRIC_PREFIX`, `PG_EXPORTER_COLLECTION_TIMEOUT`
- Datasource: `DATA_SOURCE_NAME`, `DATA_SOURCE_URI(_FILE)`, `DATA_SOURCE_USER(_FILE)`, `DATA_SOURCE_PASS(_FILE)`

Issue #1371, opened 2026-08-22
Deprecate environment variable configuration in favor of CLI flags


nicolastakashi:
**Summary**

`postgres_exporter` supports env-var configuration in two ways. We want to move fully to CLI flags, deprecate the env vars, and remove them after 2 releases.

**Current state**

1. Flags with an `Envar()` fallback, settable via flag or env var. Kingpin doesn't record which one was used: `PG_EXPORTER_WEB_TELEMETRY_PATH`, `PG_EXPORTER_DISABLE_DEFAULT_METRICS`, `PG_EXPORTER_AUTO_DISCOVER_DATABASES`, `PG_EXPORTER_EXTEND_QUERY_PATH`, `PG_EXPORTER_CONSTANT_LABELS`, `PG_EXPORTER_EXCLUDE_DATABASES`, `PG_EXPORTER_INCLUDE_DATABASES`, `PG_EXPORTER_METRIC_PREFIX`, `PG_EXPORTER_COLLECTION_TIMEOUT`. Each already has a matching CLI flag (several already marked DEPRECATED for other reasons).

2. Env-var-only settings with no CLI flag (`exporter/datasource.go:GetDataSources()`): `DATA_SOURCE_NAME`, `DATA_SOURCE_USER(_FILE)`, `DATA_SOURCE_PASS(_FILE)`, `DATA_SOURCE_URI(_FILE)`. This is the Postgres connection config, and there's no flag today for setting it.

3. The `prometheus-community/helm-charts` postgres-exporter chart sets `DATA_SOURCE_NAME`, `DATA_SOURCE_URI(_FILE)`, `DATA_SOURCE_USER(_FILE)`, and `DATA_SOURCE_PASS(_FILE)` directly as container env vars in `templates/deployment.yaml`, confirmed against the chart's current `main` branch. It has no equivalent flag-based path and will need updating in that separate repo.

ArthurSens:
Same feedback we gave to https://github.com/prometheus-community/postgres_exporter/pull/1376: let's not have sensitive information like database secrets in CLI flags. This can be easily exploited because CLI flags are written to files in the /proc filesystem.

PR #1376, opened 2026-08-23
Add --datasource.* CLI flags to replace DATA_SOURCE_* env vars

### LINKED ISSUE #1372: Add --datasource.* CLI flags to replace DATA_SOURCE_* env vars

nicolastakashi:
Closes #1372. Part of #1371.

Adds CLI flags for the Postgres connection config, which was previously only settable via env vars:

- `--datasource.dsn` (replaces `DATA_SOURCE_NAME`)
- `--datasource.uri` (replaces `DATA_SOURCE_URI`) / `--datasource.uri-file` (replaces `DATA_SOURCE_URI_FILE`)
- `--datasource.user` (replaces `DATA_SOURCE_USER`) / `--datasource.user-file` (replaces `DATA_SOURCE_USER_FILE`)
- `--datasource.pass` (replaces `DATA_SOURCE_PASS`) / `--datasource.pass-file` (replaces `DATA_SOURCE_PASS_FILE`)

`GetDataSources()` now takes a `DataSourceOpts` struct. Each field falls back to its existing env var when unset, so behavior is unchanged when only env vars are set. Precedence is unchanged: dsn wins over everything, `*_FILE` wins over the plain value, whether set by flag or env var.

Env vars still work as-is here. Deprecation warnings (#1373) and doc updates (#1374) are separate.

### DIFF
--- a/cmd/postgres_exporter/main.go
+++ b/cmd/postgres_exporter/main.go
@@ -51,6 +51,13 @@ var (
 	includeDatabases      = kingpin.Flag("include-databases", "A list of databases to include when autoDiscoverDatabases is enabled (DEPRECATED)").Default("").Envar("PG_EXPORTER_INCLUDE_DATABASES").String()
 	metricPrefix          = kingpin.Flag("metric-prefix", "A metric prefix can be used to have non-default (not \"pg\") prefixes for each of the metrics").Default("pg").Envar("PG_EXPORTER_METRIC_PREFIX").String()
 	collectionTimeout     = kingpin.Flag("collection-timeout", "Timeout for collecting the statistics when the database is slow").Default("1m").Envar("PG_EXPORTER_COLLECTION_TIMEOUT").String()
+	datasourceDSN         = kingpin.Flag("datasource.dsn", "Comma-separated list of data source names. Takes precedence over all other datasource.* flags.").Default("").String()
+	datasourceURI         = kingpin.Flag("datasource.uri", "Data source URI, in host:port/dbname form.").Default("").String()
+	datasourceURIFile     = kingpin.Flag("datasource.uri-file", "Path to a file containing the data source URI. Takes precedence over datasource.uri.").Default("").String()
+	datasourceUser        = kingpin.Flag("datasource.user", "Data source user name.").Default("").String()
+	datasourceUserFile    = kingpin.Flag("datasource.user-file", "Path to a file containing the data source user name. Takes precedence over datasource.user.").Default("").String()
+	datasourcePass        = kingpin.Flag("datasource.pass", "Data source password.").Default("").String()
+	datasourcePassFile    = kingpin.Flag("datasource.pass-file", "Path to a file containing the data source password. Takes precedence over datasource.pass.").Default("").String()
 	collectorFlags        = newCollectorFlags()
 	statStatementsFlags   = newPGStatStatementsFlags()
 	logger                = promslog.NewNopLogger()
@@ -141,7 +148,15 @@ func main() {
 		logger.Warn("Error loading config", "err", err)
 	}
 
-	dsns, err := exporter.GetDataSources()
+	dsns, err := exporter.GetDataSources(exporter.DataSourceOpts{
+		DSN:      *datasourceDSN,
+		URI:      *datasourceURI,
+		URIFile:  *datasourceURIFile,
+		User:     *datasourceUser,
+		UserFile: *datasourceUserFile,
+		Pass:     *datasourcePass,
+		PassFile: *datasourcePassFile,
+	})
 	if err != nil {
 		logger.Error("Failed reading data sources", "err", err.Error())
 		os.Exit(1)
--- a/exporter/datasource.go
+++ b/exporter/datasource.go
@@ -115,19 +115,38 @@ func (e *Exporter) scrapeDSN(ch chan<- prometheus.Metric, dsn string) error {
 	return server.Scrape(ch)
 }
 
+// DataSourceOpts holds the values of the --datasource.* flags. An empty
+// field falls back to the corresponding DATA_SOURCE_* environment variable.
+type DataSourceOpts struct {
+	DSN      string
+	URI      string
+	URIFile  string
+	User     string
+	UserFile string
+	Pass     string
+	PassFile string
+}
+
+func flagOrEnvVar(flagVal, envKey string) string {
+	if flagVal != "" {
+		return flagVal
+	}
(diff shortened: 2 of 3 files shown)


sysadmind:
Okay this one I'm less sure about. Specifically anywhere that we take in values that can be considered secret, I am against using a CLI flag. It is too easy to expose those values when they are passed via flag.

nicolastakashi:
> Okay this one I'm less sure about. Specifically anywhere that we take in values that can be considered secret, I am against using a CLI flag. It is too easy to expose those values when they are passed via flag.

I agree 100%, I was thinking about introduce a --datasource.config-file that people could use one or another and config files could be mounted from secrets, wdyt @sysadmind ?

ArthurSens:
Agree that a config file is safer than CLI flags for sensitive information! I'm just not sure if we should create yet another configuration file or if we should extend the existing auth-modules file 🤔

sysadmind:
I think we should combine with auth modules if we're looking at a file for config. That said, I think we should keep env vars for secrets. Like you might choose to pass `DATASOURCE_PASSWORD` or something similar for the password to connect. That makes something like passing it to a pod in Kubernetes from a secret easy. Kubernetes is another great reason not to have sensitive values in cli flags because it's so easily readable in the API.

ArthurSens:
I'm not a fan of env vars, it also breaks the embedded use case. One single binary running several instances of Postgres exporters cannot set multiple env vars for each secret. I'd rather keep everything on files 😬

sysadmind:
It only breaks embedding if that's in the part of the library that's embedded. That's what the cmd/ or main package is for. It hooks up things for THIS cli binary.

ArthurSens:
good point!

nicolastakashi:
I personally would avoid two ways of configuring secrets, but I think we need to be all or nothing.

So I'm fine on keeping those two options as env vars and expanding the auth modules later.

I'll patch the PR with that change

ArthurSens:
Friendly ping to let you know we're still waiting for the changes to move secrets away from the CLI. I don't mean to rush here, just sending the message as I'm triaging open PRs :)

nicolastakashi:
PTAL @ArthurSens and @sysadmind sorry for the delay on getting back to this.

cmd/postgres_exporter/main.go
```diff
@@ -52,6 +52,10 @@ var (
 	metricPrefix          = kingpin.Flag("metric-prefix", "A metric prefix can be used to have non-default (not \"pg\") prefixes for each of the metrics").Default("pg").Envar("PG_EXPORTER_METRIC_PREFIX").String()
 	collectionTimeout     = kingpin.Flag("collection-timeout", "Timeout for collecting the statistics when the database is slow").Default("1m").Envar("PG_EXPORTER_COLLECTION_TIMEOUT").String()
 	wrapLargeCounters     = kingpin.Flag("wrap-large-counters", "Wrap 64-bit counters at 2^53 to avoid floating point rounding.").Default("true").Bool()
+	datasourceURI         = kingpin.Flag("datasource.uri", "Data source URI, in host:port/dbname form.").Default("").String()
+	datasourceURIFile     = kingpin.Flag("datasource.uri-file", "Path to a file containing the data source URI. Takes precedence over datasource.uri.").Default("").String()
+	datasourceUserFile    = kingpin.Flag("datasource.user-file", "Path to a file containing the data source user name.").Default("").String()
+	datasourcePassFile    = kingpin.Flag("datasource.pass-file", "Path to a file containing the data source password.").Default("").String()
```
ArthurSens, line 58: Based on our meeting notes from last Friday, I believe we agreed not to use files for anything in "Single target" mode. We're also happy with an environment variable for the password. Maybe `user` is also sensitive data, so we can keep it that way as well.

What do you think @sysadmind?

ArthurSens (commented):
Following the architecture we've set based on https://github.com/prometheus/prometheus-opentelemetry-collector/blob/main/docs/embeddable-exporters.md, I think `cmd/postgres_exporter/main.go` is probably the best place to read CLI/Env vars and wire it to the `Config` package. We've been trying to deprecate and remove as much code as possible from the `exporter` pkg, so maybe we could move things to cmd instead?


Also, this PR is a great opportunity to add deprecation notes to the changelog! Could we add some notes listing all the deprecated env vars and how the new CLI flags are going to replace them?

cmd/postgres_exporter/main.go
sysadmind: I'm not opposed to having the file for a password. In general though, I don't want to remove environment variables for sensitive values. I think these are fine as is, so long as we don't remove the env var for password (I'm okay with username too).

cmd/postgres_exporter/main.go
```diff
@@ -52,6 +52,10 @@ var (
 	metricPrefix          = kingpin.Flag("metric-prefix", "A metric prefix can be used to have non-default (not \"pg\") prefixes for each of the metrics").Default("pg").Envar("PG_EXPORTER_METRIC_PREFIX").String()
 	collectionTimeout     = kingpin.Flag("collection-timeout", "Timeout for collecting the statistics when the database is slow").Default("1m").Envar("PG_EXPORTER_COLLECTION_TIMEOUT").String()
 	wrapLargeCounters     = kingpin.Flag("wrap-large-counters", "Wrap 64-bit counters at 2^53 to avoid floating point rounding.").Default("true").Bool()
+	datasourceURI         = kingpin.Flag("datasource.uri", "Data source URI, in host:port/dbname form.").Default("").String()
```
sysadmind, line 55: This doesn't seem like it's hooked up to collector/ at all? The exporter/ code is just the legacy style collectors, but we need to have all of this hooked up to the collect/ package too with its runtime.

sysadmind (changes requested):

PR #1377, opened 2026-08-23
Docs fix: Update mixin/README for go 1.17+

zoeimogen:
Using `go get` to install binaries was deprecated in 1.17, and removed in 1.18 - `go install` is now required to build the binaries. See https://go.dev/doc/go1.18#go-command for more details.


ArthurSens (approved):
Thanks!

PR #533, opened 2021-05-17
pg_process_idle should be called only once database in instance

v0112358:
If `auto-discover-databases` is enabled, postgres_exporter will return 500 status.
- postgres_exporter version 0.9.0.
- The error message:
```
....
* collected metric "pg_process_idle_seconds" { label:<name:"application_name" value:"PostgreSQL JDBC Driver" > label:<name:"server" value:"127.0.0.1:5432" > histogram:<sample_count:5 sample_sum:1256 bucket:<cumulative_count:0 upper_bound:1 > bucket:<cumulative_count:0 upper_bound:2 > bucket:<cumulative_count:0 upper_bound:5 > bucket:<cumulative_count:0 upper_bound:15 > bucket:<cumulative_count:0 upper_bound:30 > bucket:<cumulative_count:1 upper_bound:60 > bucket:<cumulative_count:2 upper_bound:90 > bucket:<cumulative_count:2 upper_bound:120 > bucket:<cumulative_count:4 upper_bound:300 > > } was collected before with the same name and label values
* collected metric "pg_process_idle_seconds" { label:<name:"application_name" value:"vertx-pg-client" > label:<name:"server" value:"127.0.0.1:5432" > histogram:<sample_count:4 sample_sum:596 bucket:<cumulative_count:0 upper_bound:1 > bucket:<cumulative_count:0 upper_bound:2 > bucket:<cumulative_count:0 upper_bound:5 > bucket:<cumulative_count:0 upper_bound:15 > bucket:<cumulative_count:0 upper_bound:30 > bucket:<cumulative_count:0 upper_bound:60 > bucket:<cumulative_count:0 upper_bound:90 > bucket:<cumulative_count:0 upper_bound:120 > bucket:<cumulative_count:4 upper_bound:300 > > } was collected before with the same name and label values
```

Issue #1326, opened 2026-06-03
Adapt new collector package for database scoped collectors


sysadmind:
`--auto-discover-databases` is deprecated. This is because it's designed to support the dynamic SQL metrics which is also deprecated. It did however provide value for certain collectors that provide metrics only for the current database. 

The new collector/ subsystem only supports a single database connection per exporter instance. NewPostgresCollector takes a single dsn string (not a slice).

  The old exporter/ subsystem with --auto-discover-databases would:
  1. Connect to your configured DSN
  2. Query pg_database for all databases on the server
  3. Open a separate connection to each database and scrape it

  This matters because several PostgreSQL system views are per-database. So a user who previously ran one exporter pointed at their postgres database with --auto-discover-databases would get table/index/lock stats for all their application databases. In the new  collector pattern, they only get those stats for the single database in their DSN.

# Proposal

I am proposing that we add the functionality back to address the user issues and restore that level of metrics collection for the end users. Below is an outline of how it could work so that we can review and decide on the direction. 

###  Update collector registration to specify cluster or database scoped

```golang
  const (
      ClusterScoped  = false
      DatabaseScoped = true
  )

  func registerCollector(name string, isDefaultEnabled bool, isDatabaseScoped bool, createFunc func(collectorConfig) (Collector, error))
```

Store isDatabaseScoped in a parallel map (like collectorState). All existing registrations default to ClusterScoped; the four per-DB collectors flip to DatabaseScoped.

### Add a discovery Option to PostgresCollector

```golang
  func WithDatabaseDiscovery(include, exclude []string) Option
```
When set, PostgresCollector stores the filter lists alongside the primary instance.

### Split Collect() by scope

  When discovery is enabled, `Collect()` becomes:
  1. Run cluster-scoped collectors once on the primary instance (current behavior)
  2. Query pg_database on the primary connection, filter by include/exclude
  3. For each discovered DB: open a new instance pointing at that DB, run only database-scoped collectors, close the instance

Without discovery (default), behavior is identical to today — all collectors run against the single instance, including the database-scoped ones (which scrape the connected DB only, as  now).

  Key properties of this design:
  - Zero behavior change for existing users who don't opt in
  - No changes to the Collector interface — individual collectors don't need to know about discovery
  - The deduplication problem is solved structurally: cluster-scoped metrics are emitted once, per-DB metrics are emitted once per database with the datname label already in place
[...]

PR #1378, opened 2026-08-25
Run database-scoped collectors against every discovered database

mwimpelberg28:
collector.NewRuntime only ever built one PostgresCollector, bound to DataSourceNames[0]. This meant per-database collectors like stat_user_tables, statio_user_tables and statio_user_indexes silently never ran for any database beyond the first configured DSN, whether that list came from a static, comma-separated DATA_SOURCE_NAME or from --auto-discover-databases.

Naively running every collector against every discovered database isn't safe either: most collectors (bgwriter, wal, stat_activity, stat_statements, locks, ...) read catalog views or functions that already report on the whole instance from any single connection, so running them again from a second database would re-report the same rows and the registry would panic on the duplicate metric (as already happened once, for stat_activity, in #533).

This adds an explicit scope to each registered collector (clusterScope vs databaseScope) and a multiInstanceCollector that fans the database-scoped ones out across every DSN exporter.Exporter currently targets (static list or, with autodiscovery enabled, freshly enumerated databases), while cluster-scoped collectors still only ever run once, against the originally configured DSN. The target list is re-evaluated on every scrape, so databases created or dropped after startup are picked up without restarting the exporter, matching the autodiscovery behavior already used by the legacy exporter.Exporter metrics.

Also gives statio_user_indexes a datname label, matching its sibling per-database collectors: without it, index metrics from two databases with a same-named index would collide once run concurrently.


mwimpelberg28:
Hi @SuperQ @wrouesnel @sysadmind — this has been open for a bit with no review yet. Also flagging that the `CI` and `golangci-lint` workflow runs are stuck in `action_required` since this is from a fork, so they need a maintainer approval to actually run.

Summary of the change: `collector.NewRuntime` previously only ever built one `PostgresCollector` bound to `DataSourceNames[0]`, so per-database collectors (`stat_user_tables`, `statio_user_tables`, `statio_user_indexes`) silently never ran for any database beyond the first configured DSN. This PR adds an explicit scope to each registered collector (`clusterScope` vs `databaseScope`) and a `multiInstanceCollector` that fans database-scoped collectors out across every DSN the exporter currently targets (static list, or freshly enumerated databases when `--auto-discover-databases` is set), while cluster-scoped collectors still run once against the originally configured DSN. It also adds a `datname` label to `statio_user_indexes` to avoid collisions across databases.

Happy to address any feedback. Thanks for taking a look!

sysadmind (changes requested):
Thanks for the PR. I think this is heading in the right direction, but it's going to need some changes to get landed. #1326 outlines the plan for database-scoped collectors. 

There are 2 main issues that I have with this PR as it is. The first is the term `cluster`. As was commented on that issue, PostgreSQL doesn't really have a cluster, so we should use `server` instead. The other issue is this new multiInstanceCollector (and all of the changes that go into making it). As it exists, I think this actually causes a lot of problems for existing users. If you try to scrape all the server scoped metrics from both the primary and a database scoped DSN, you create duplicate metrics which makes the registry really mad.  

The strategy that I outlined in #1326 is to always collect the server scoped items, as is the behavior today, and then run database scoped collectors over the databases. That feels like a much better strategy to me. We need to go that direction unless there's a really good reason to deviate.

mwimpelberg28:
Thanks for the review, @sysadmind — pushed a fix that follows the strategy from #1326.

**Terminology**: renamed `clusterScope` to `serverScope` throughout (PostgreSQL doesn't have a "cluster").

**Removed `multiInstanceCollector`**: it built the full server+database collector set for *every* explicitly configured DSN, which panics the registry once two of those DSNs turn out to be the same physical server (e.g. two databases on one server, each listed separately in `DATA_SOURCE_NAME`).

**New design**: exactly one primary connection (`DataSourceNames[0]`) ever runs server-scoped collectors, so they can never be duplicated no matter how many databases exist. `PostgresCollector` gains a `WithDatabaseDiscovery` option that, when `--auto-discover-databases` is set, queries `pg_database` on that same primary connection and runs only the database-scoped collectors against every other database found, each over its own connection, reconciled on every scrape. This also drops the collector package's dependency on `exporter.Target`/`TargetDSNs` — discovery is now entirely self-contained in `collector/database_discovery.go`.

Also fixed the DCO check, which was failing on the previous push. `CI` and `golangci-lint` are still sitting in `action_required` since this is a fork PR — those need a maintainer to approve the workflow run before they'll execute.

collector/database_discovery.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// connstringRe superficially validates a key=value DSN: connstring syntax is
+// complex (and not sure if even regular), and we don't need to parse it, so
+// we just check that it starts with a valid-ish keyword pair.
+var connstringRe = regexp.MustCompile(`^ *[a-zA-Z0-9]+ *= *[^= ]+`)
+
+// databaseDSN returns dsn rewritten to connect to database instead of
+// whichever database it originally pointed at.
+func databaseDSN(dsn, database string) (string, error) {
```
sysadmind: We have a config package that has a DSN object. It also nicely handles redaction of passwords when we print a string. I don't know that we should duplicate any connection string logic here. 

Also, I think that the instance type might be the right place for getting a new connection to a specific database.

collector/database_discovery.go
```diff
@@ -0,0 +1,166 @@
...
+			continue
+		}
+		databases = append(databases, database)
+	}
+	return databases, rows.Err()
+}
+
+// collect reconciles the cached per-database collectors against the
+// databases currently on the server, then scrapes every one of them.
+func (d *databaseDiscovery) collect(ctx context.Context, primaryDSN string, db *sql.DB, ch chan<- prometheus.Metric, logger *slog.Logger, build func(dsn string) (*PostgresCollector, error)) {
```
sysadmind: Why do we need a separate collect func here? This seems overly complicated. Why not update Collect or collectFromConnection on PostgresCollector?

collector/collector.go
```diff
@@ -160,6 +192,53 @@ func WithCollectorStates(states map[string]bool) Option {
...
+		p.databaseDiscovery = newDatabaseDiscovery(includeDatabases, excludeDatabases)
+		return nil
+	}
+}
+
+// buildDatabaseCollector creates a PostgresCollector that only runs
+// database-scoped collectors against dsn, sharing every other setting with
+// p. It is used to scrape a database discovered alongside p's own primary
+// connection.
+func (p *PostgresCollector) buildDatabaseCollector(dsn string) (*PostgresCollector, error) {
```
sysadmind: I'm not sure why we need this. The existing PostgresCollector already knows about database or server scoped collectors. Why do we need to duplicate the collector?

collector/pg_statio_user_indexes.go
```diff
@@ -67,12 +68,16 @@ func (c *PGStatioUserIndexesCollector) Update(ctx context.Context, instance *ins
...
 	for rows.Next() {
-		var schemaname, relname, indexrelname sql.NullString
+		var datname, schemaname, relname, indexrelname sql.NullString
 		var idxBlksRead, idxBlksHit sql.NullInt64
 
-		if err := rows.Scan(&schemaname, &relname, &indexrelname, &idxBlksRead, &idxBlksHit); err != nil {
+		if err := rows.Scan(&datname, &schemaname, &relname, &indexrelname, &idxBlksRead, &idxBlksHit); err != nil {
 			return err
 		}
+		datnameLabel := "unknown"
```
sysadmind: This is really brittle. If we have 2 databases that have invalid datname columns, they're both "unknown" and now present a conflict in the metric registry.

sysadmind (changes requested):
I think that the naming changes are good. I don't think that we want to create all of these duplicate collectors though. A single PostgresCollector already knows about which collectors are server or database scoped. We can just run the server scoped collectors first, build the database list, and run the database scoped collectors for each of those. 

The current implementation still returns errors for the metrics endpoint which makes me question if this PR is being tested locally. Please ensure that this works with a local postgres server with multiple databases before asking for another review:
```
$ curl -w "HTTP %{http_code}" localhost:19187/metrics
HTTP 500
4 error(s) occurred:
* collected metric "pg_scrape_collector_duration_seconds" { label:{name:"collector" value:"stat_user_tables"} ...} was collected before with the same name and label values
* collected metric "pg_scrape_collector_success" { label:{name:"collector" value:"stat_user_tables"} ...} was collected before with the same name and label values
* ...
```

mwimpelberg28:
Thanks for the thorough review, @sysadmind. I've pushed changes addressing all four inline comments plus the remaining feedback from your second review (the duplicate-collector concern and the request to test against a local Postgres before asking for another look).

## Inline comments

**`database_discovery.go:59` — duplicated DSN parsing, instance is the right place for this**

Removed `databaseDSN` and the duplicated `connstringRe` regex entirely. `config.DSN` now has a `WithDatabase(database string) DSN` method (built on the existing `dsnFromString`/`ParseDSN` parsing, so no logic is duplicated), and `instance` gained a `withDatabase(database string) (*instance, error)` method that uses it to build a connection to a different database on the same server.

**`database_discovery.go:105` — why a separate `collect` func instead of updating `Collect`/`collectFromConnection`?**

You were right that this was overcomplicated. The whole `databaseDiscovery` struct (its cache of `*PostgresCollector`, mutex, and its own `collect`/`Close` methods) is gone. `collectFromConnection` now takes an optional scope filter, so it can run either every collector (primary connection) or just the database-scoped ones (discovered connections) against a given `*instance`, using the single `p.Collectors` map both times. `PostgresCollector.Collect` calls it directly, then a small `collectDiscoveredDatabases` helper discovers other databases and fans out to them — no second collector, no persisted cache. Turns out no caching was needed at all: the primary connection already opens and closes a fresh `sql.DB` on every scrape (`Collect` does `copy → setup → defer Close`), so there was never a live connection to actually preserve across scrapes for the discovered databases either.

**`collector.go:227` — why duplicate the collector instead of reusing scope info `PostgresCollector` already has?**

Same fix as above — `buildDatabaseCollector` and `onlyScope`/`scopeFilter` are deleted. There is exactly one `PostgresCollector` and one `Collectors` map now; discovered databases are just additional `*instance` connections that the same collector set runs against, filtered by scope at collect-time.

**`pg_statio_user_indexes.go:77` — brittle `"unknown"` datname fallback**

`datname` is scanned as a plain, non-nullable `string` now instead of `sql.NullString` with an `"unknown"` fallback. It comes from `current_database()`, which Postgres guarantees is never NULL, so there's no longer any path that could produce a shared sentinel value across databases — which was the actual collision risk once this collector started running concurrently against multiple databases.

## Second review — "duplicate collectors" and the local-testing request

The redesign above (single `PostgresCollector`, scope-filtered `collectFromConnection`) is the same change you described: run server-scoped collectors once on the primary connection, build the database list, run database-scoped collectors against each.

On testing locally — you were right to push on this, and it caught a real bug. While reworking the above I found that `execute()` emits `pg_scrape_collector_duration_seconds{collector="<name>"}` and `pg_scrape_collector_success{collector="<name>"}` with no `datname` label, so once a database-scoped collector like `stat_user_tables` runs against more than one database in a scrape, those two meta-metrics collide on identical labels — the exact class of error from your repro:
```
* collected metric "pg_scrape_collector_duration_seconds" { label:{name:"collector" value:"stat_user_tables"} ...} was collected before with the same name and label values
```
Fixed by adding a `datname` label to both descriptors, sourced from a new `instance.database` field populated via `SELECT current_database()` during `setup()` (the same guaranteed-non-null source used for the `pg_statio_user_indexes` fix above).

## Testing

Ran this against a real local Postgres (Docker, `postgres:16`) with three databases (`postgres`, `dbtwo`, `dbthree`), each with a table + its implicit index, using `--auto-discover-databases`:

```
$ curl -w "HTTP %{http_code}" localhost:19187/metrics
HTTP 200
```
(previously HTTP 500 with the duplicate-metric errors from your repro)

`pg_exporter_last_scrape_error 0`, and the database-scoped data is correctly split per database:
```
pg_stat_user_tables_seq_scan{datname="dbthree",relname="t3",schemaname="public"} 1
pg_stat_user_tables_seq_scan{datname="dbtwo",relname="t2",schemaname="public"} 1
pg_stat_user_tables_seq_scan{datname="postgres",relname="t1",schemaname="public"} 1
```
and the previously-colliding meta-metrics are now disambiguated:
```
pg_scrape_collector_duration_seconds{collector="stat_user_tables",datname="dbthree"} 0.0028
pg_scrape_collector_duration_seconds{collector="stat_user_tables",datname="dbtwo"} 0.0024
pg_scrape_collector_duration_seconds{collector="stat_user_tables",datname="postgres"} 0.0163
```
With `--collector.statio_user_indexes` also enabled, confirmed the datname fix directly — real, distinct database names, no `"unknown"`:
```
pg_statio_user_indexes_idx_blks_hit_total{datname="dbthree",indexrelname="t3_pkey",relname="t3",schemaname="public"} 0
pg_statio_user_indexes_idx_blks_hit_total{datname="dbtwo",indexrelname="t2_pkey",relname="t2",schemaname="public"} 0
pg_statio_user_indexes_idx_blks_hit_total{datname="postgres",indexrelname="t1_pkey",relname="t1",schemaname="public"} 0
```
Also verified the "reconciled on every scrape" behavior: dropping `dbthree` mid-run made its metrics disappear on the very next scrape with no exporter restart and no error, and creating a brand-new `dbfour` made its metrics appear on the next scrape.

`go build`, `go vet` (including `-tags integration`), `go test ./...`, and `go test -race ./collector/...` all pass. Added unit coverage for `config.DSN.WithDatabase`, `instance.withDatabase`, and the scope-filtering behavior of `collectFromConnection`, plus an integration test (`-tags integration`) that registers the collector with a real `prometheus.Registry` and calls `Gather()` against a discovered database, so a future regression of the duplicate-metric issue would fail CI rather than only showing up against a live server.

collector/collector.go
```diff
@@ -214,20 +271,67 @@ func (p PostgresCollector) Collect(ch chan<- prometheus.Metric) {
...
+	defer cancel()
+
+	databases, err := discoverDatabases(ctx, primary.getDB(), p.databaseDiscovery.includeDatabases, p.databaseDiscovery.excludeDatabases)
+	if err != nil {
+		p.logger.Error("failed to discover databases", "err", err)
+		return
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(len(databases))
```
ArthurSens, line 297: I wonder if we should make the concurrent worker size configurable. I can imagine we exhausting database connections in a server with hundreds/thousands of databases, but it would also become a very slow if we do it sequentially.

collector/pg_statio_user_indexes.go
```diff
@@ -36,18 +36,19 @@ var (
...
 		"Number of disk blocks read from this index",
-		[]string{"schemaname", "relname", "indexrelname"},
+		[]string{"datname", "schemaname", "relname", "indexrelname"},
 		prometheus.Labels{},
 	)
 	statioUserIndexesIdxBlksHit = prometheus.NewDesc(
 		prometheus.BuildFQName(namespace, statioUserIndexesSubsystem, "idx_blks_hit_total"),
 		"Number of buffer hits in this index",
-		[]string{"schemaname", "relname", "indexrelname"},
+		[]string{"datname", "schemaname", "relname", "indexrelname"},
```
ArthurSens, line 45: Adding a new label unconditionally breaks behavior for users who do not enable this new feature. 

We should only add `datname` if the user explicitly enables it!

collector/runtime.go
```diff
@@ -47,16 +47,28 @@ func NewRuntime(validatedConfig config.ValidatedConfig, logger *slog.Logger) (*R
...
+		// database on the server too. Server-scoped collectors are
+		// unaffected: they still only ever run once, against
+		// DataSourceNames[0], so they are never duplicated.
+		opts = append(opts, WithDatabaseDiscovery(cfg.IncludeDatabases, cfg.ExcludeDatabases))
+	}
+
+	postgresCollector, err := NewPostgresCollector(
+		logger,
+		cfg.ExcludeDatabases,
+		cfg.DataSourceNames[0],
```
ArthurSens, line 69: I'm a bit confused about us still sending `DataSourceNames[0]` here. Doesn't that mean that if somebody passes several servers, the database-scoped collector will only run for the first server?

I'm not sure my interpretation is correct, so I'm happy to be corrected.

ArthurSens: If my interpretation is correct, the best way forward may be to stop pretending that DataSourceNames accepts multiple DSNs and ensure it accepts only one.

The multi-target pattern is the architecture we want to enable for multi-server use cases.

collector/runtime.go
```diff
@@ -47,16 +47,28 @@ func NewRuntime(validatedConfig config.ValidatedConfig, logger *slog.Logger) (*R
...
+		// database on the server too. Server-scoped collectors are
+		// unaffected: they still only ever run once, against
+		// DataSourceNames[0], so they are never duplicated.
+		opts = append(opts, WithDatabaseDiscovery(cfg.IncludeDatabases, cfg.ExcludeDatabases, cfg.AutoDiscoverDatabasesMaxConcurrency))
+	}
+
+	postgresCollector, err := NewPostgresCollector(
+		logger,
+		cfg.ExcludeDatabases,
+		cfg.DataSourceNames[0],
```
mwimpelberg28: Confirmed: `DataSourceNames[0]` here is unchanged from master (predates this PR by several releases — it was already the case for the native `PostgresCollector` before this PR added database discovery). This PR only fans database-scoped collectors out across databases discovered on *one* server; it does not change how multiple, separately-configured servers are handled by the native collector path.

Given the two suggested directions here point opposite ways (build full multi-DSN fan-out with per-server dedup vs. restrict `DataSourceNames` to a single DSN and lean on the existing `/probe` multi-target pattern), I do not want to bundle that architecture decision into this PR. Filing this as a separate follow-up so it can get its own discussion/PR. Left the discovery-specific fixes (bounded concurrency, context propagation, socket DSN serialization, datname label gating) in the latest push.

mwimpelberg28:
@/private/tmp/claude-501/-Users-mwimpelberg-git-postgres-exporter/fedae9d2-9ff0-47f2-8292-6ec3ebd82bfa/scratchpad/pr1378_comment.md

collector/collector.go
mwimpelberg28: Fixed: added an `AutoDiscoverDatabasesMaxConcurrency` config field (`--auto-discover-databases.max-concurrency`, default 10) and bounded `collectDiscoveredDatabases` with a semaphore sized by it, so a server with many databases cannot open more connections than configured.

mwimpelberg28: Fixed: added an `AutoDiscoverDatabasesMaxConcurrency` config field (`--auto-discover-databases.max-concurrency`, default 10) and bounded `collectDiscoveredDatabases` with a semaphore sized by it, so a server with many databases cannot open more connections than configured.

collector/pg_statio_user_indexes.go
mwimpelberg28: Fixed: `pg_statio_user_indexes`'s `datname` label is now only added when `WithDatabaseDiscovery` is active (i.e. `--auto-discover-databases` is set) — that's the only case where this collector can legitimately run concurrently against more than one database and need it to disambiguate. Installs that never enable discovery see no schema change on this metric. Applied the same gating to the `pg_scrape_collector_duration_seconds`/`_success` meta-metrics, which got the same unconditional label last round for the same underlying reason.

collector/runtime.go
mwimpelberg28: Confirmed — this line is unchanged from master and predates this PR (the native `PostgresCollector` already only used `DataSourceNames[0]` before this PR added database discovery). This PR only fans database-scoped collectors out across databases discovered on *one* server; it does not change how multiple, separately-configured servers are handled.

Your suggestion here and Copilot's comment below point in different directions (full multi-DSN fan-out with per-server dedup vs. restricting `DataSourceNames` to one DSN and leaning on `/probe`), so I'd rather not bundle that architecture decision into this PR. Filing it as a follow-up issue instead — the fixes actually in scope for this PR (bounded concurrency, context propagation, socket DSN serialization, datname label gating) are in the latest push.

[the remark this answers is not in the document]
mwimpelberg28: Same as my reply above to ArthurSens's comment on this line: confirmed `DataSourceNames[0]` is unchanged from master and predates this PR. It's a pre-existing limitation of the native collector framework, not something this PR changes — this PR only fans database-scoped collectors out across databases discovered on one server. Filing a follow-up issue for proper multi-server support in the native collector path rather than bundling it here.

config/dsn.go
```diff
@@ -62,6 +62,26 @@ func (d DSN) GetConnectionString() string {
...
 }
 
+// WithDatabase returns a copy of d pointed at database instead of whichever
+// database it originally targeted. The override is always carried in the
+// query string (as "dbname"), regardless of how the database was originally
+// specified, since a "dbname" left behind in the query string would
+// otherwise take precedence over the path when the driver parses the
+// resulting connection string.
+func (d DSN) WithDatabase(database string) DSN {
+	d.path = ""
```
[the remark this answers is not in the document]
mwimpelberg28: Confirmed and fixed — thanks for catching this. `pq`'s own doc comment on `convertURL` confirms Go's `net/url.Parse` rejects a percent-encoded `/` in the host, so `dsn.WithDatabase(...).GetConnectionString()` on a `host=/var/run/postgresql/` primary DSN (the form documented in this project's own README) produced a string `sql.Open` can't even parse — verified this fails with "invalid URL escape" before the fix.

`GetConnectionString` now detects a socket-shaped host (leading `/` or `@`, matching the `host` keyword's documented rules) and serializes as key=value instead of a URL, the only form `pq` accepts for a socket path. Added regression tests and verified the output round-trips through the driver's own parser.

collector/collector.go
```diff
@@ -214,20 +271,67 @@ func (p PostgresCollector) Collect(ch chan<- prometheus.Metric) {
...
+	for _, database := range databases {
+		go func(database string) {
+			defer wg.Done()
```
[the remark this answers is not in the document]
mwimpelberg28: Fixed: added an `AutoDiscoverDatabasesMaxConcurrency` config field (`--auto-discover-databases.max-concurrency`, default 10) and bounded the goroutines in `collectDiscoveredDatabases` with a semaphore sized by it, so a scrape can never open more discovered-database connections at once than configured (same fix as ArthurSens' comment above on this file).

collector/collector.go
```diff
@@ -214,20 +271,67 @@ func (p PostgresCollector) Collect(ch chan<- prometheus.Metric) {
...
+
+			inst, err := primary.withDatabase(database)
+			if err != nil {
+				p.logger.Error("failed to build connection for discovered database", "database", database, "err", err)
+				return
+			}
+			defer inst.Close()
+			if err := inst.setup(); err != nil {
```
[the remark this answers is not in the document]
mwimpelberg28: Fixed: `instance.setup` and `queryVersion` now take a `context.Context`. Each discovered database's goroutine builds one `context.WithTimeout(context.Background(), p.CollectionTimeout)` covering both `setup()` (connect + version + `current_database()`) and the subsequent `collectFromConnection` call, so a stalled connect/auth is now bounded by `CollectionTimeout` instead of blocking `wg.Wait()` indefinitely. Applied the same fix to the primary connection's `setup()` call in `Collect`, which had the identical gap.

collector/pg_statio_user_indexes.go
ArthurSens: Hmmm, thinking a bit more about this... it feels weird that the `datname` label is added conditionally to this metric but not to others.

I'd love to see consistency somehow, and I'm afraid we'll need to introduce a breaking change to achieve it. Either we remove the `datname` label from all metrics and make it available only behind the flag you've introduced, or we add `datname` everywhere unconditionally. In both cases, we're breaking people.

For the long term, I think it's better not to add them! We introduced the multi-target pattern specifically to monitor several databases with a single postgres_exporter, and with it, it's much easier to add the `datname` label when needed through relabeling.

I'd love to see @sysadmind opinion on this

collector/collector.go
```diff
@@ -207,27 +310,88 @@ func (p PostgresCollector) Collect(ch chan<- prometheus.Metric) {
...
+			defer wg.Done()
+
+			sem <- struct{}{}
+			defer func() { <-sem }()
+
+			// One deadline covers both connecting to database and
+			// scraping it, so a stalled connect/authentication cannot
+			// keep the scrape blocked past CollectionTimeout.
+			ctx, cancel := context.WithTimeout(context.Background(), p.CollectionTimeout)
+			defer cancel()
```
ArthurSens: I think this is incorrect, since we're now processing things in batches; each batch gets a fresh collection timeout.

Let's say we have a 20s collection timeout, 1000 databases, and 10 workers. Let's also say each batch takes 5 seconds. The first, second, third, and fourth batches are all under the 20s timeout, but together they hit 20s, so they all succeed without problems. When the fifth batch starts, the timeout would have already passed, but since we're measuring it against a `context.Background()` with a 20s timeout again, nothing stops it.

This will lead to a 500s collection that won't stop within the first 20s of the collection timeout.


here's my attempt at fixing it:
```diff
diff --git a/collector/collector.go b/collector/collector.go
index c126aee..d922933 100644
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -324,7 +324,7 @@ func (p PostgresCollector) Collect(ch chan<- prometheus.Metric) {
 	p.collectFromConnection(ctx, inst, ch, nil)
 
 	if p.databaseDiscovery != nil {
-		p.collectDiscoveredDatabases(inst, ch)
+		p.collectDiscoveredDatabases(ctx, inst, ch)
 	}
 }
 
@@ -335,48 +335,30 @@ func (p PostgresCollector) Collect(ch chan<- prometheus.Metric) {
 // are picked up without a restart; no state is kept between calls. At most
 // databaseDiscovery.maxConcurrency databases are scraped at once, so a
 // server with many databases cannot make a single scrape exhaust its
// connection slots.
-func (p PostgresCollector) collectDiscoveredDatabases(primary *instance, ch chan<- prometheus.Metric) {
-	discoveryCtx, cancel := context.WithTimeout(context.Background(), p.CollectionTimeout)
-	defer cancel()
-
-	databases, err := discoverDatabases(discoveryCtx, primary.getDB(), p.databaseDiscovery.includeDatabases, p.databaseDiscovery.excludeDatabases)
+func (p PostgresCollector) collectDiscoveredDatabases(ctx context.Context, primary *instance, ch chan<- prometheus.Metric) {
+	databases, err := discoverDatabases(ctx, primary.getDB(), p.databaseDiscovery.includeDatabases, p.databaseDiscovery.excludeDatabases)
 	if err != nil {
 		p.logger.Error("failed to discover databases", "err", err)
 		return
 	}
 
-	sem := make(chan struct{}, p.databaseDiscovery.maxConcurrency)
-	var wg sync.WaitGroup
-	wg.Add(len(databases))
-	for _, database := range databases {
-		go func(database string) {
-			defer wg.Done()
-
-			sem <- struct{}{}
-			defer func() { <-sem }()
-
-			// One deadline covers both connecting to database and
-			// scraping it, so a stalled connect/authentication cannot
-			// keep the scrape blocked past CollectionTimeout.
-			ctx, cancel := context.WithTimeout(context.Background(), p.CollectionTimeout)
-			defer cancel()
-
-			inst, err := primary.withDatabase(database)
-			if err != nil {
-				p.logger.Error("failed to build connection for discovered database", "database", database, "err", err)
-				return
-			}
-			defer inst.Close()
-			if err := inst.setup(ctx); err != nil {
-				p.logger.Error("failed to connect to discovered database", "database", database, "err", err)
-				return
-			}
-			scope := databaseScope
-			p.collectFromConnection(ctx, inst, ch, &scope)
-		}(database)
-	}
-	wg.Wait()
+	forEachDatabase(ctx, databases, p.databaseDiscovery.maxConcurrency, func(ctx context.Context, database string) {
+		inst, err := primary.withDatabase(database)
+		if err != nil {
+			p.logger.Error("failed to build connection for discovered database", "database", database, "err", err)
+			return
+		}
+		defer inst.Close()
+		if err := inst.setup(ctx); err != nil {
+			p.logger.Error("failed to connect to discovered database", "database", database, "err", err)
+			return
+		}
+		scope := databaseScope
+		p.collectFromConnection(ctx, inst, ch, &scope)
+	})
 }
 
 // collectFromConnection runs the Collectors matching scope against inst,
diff --git a/collector/database_discovery.go b/collector/database_discovery.go
index 4366a8f..757758e 100644
--- a/collector/database_discovery.go
+++ b/collector/database_discovery.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"database/sql"
 	"slices"
+	"sync"
 )
 
 // discoverDatabases returns every database on the server db is connected to,
@@ -45,3 +46,42 @@ func discoverDatabases(ctx context.Context, db *sql.DB, includeDatabases, exclud
 	}
 	return databases, rows.Err()
 }
+
+func forEachDatabase(ctx context.Context, databases []string, maxConcurrency int, collect func(context.Context, string)) {
+	if len(databases) == 0 {
+		return
+	}
+	if maxConcurrency <= 0 {
+		maxConcurrency = 1
+	}
+	if maxConcurrency > len(databases) {
+		maxConcurrency = len(databases)
+	}
+
+	jobs := make(chan string)
+	var wg sync.WaitGroup
+	wg.Add(maxConcurrency)
+	for range maxConcurrency {
+		go func() {
+			defer wg.Done()
+			for database := range jobs {
+				if ctx.Err() != nil {
+					return
+				}
+				collect(ctx, database)
+			}
+		}()
+	}
+
+	for _, database := range databases {
+		select {
+		case jobs <- database:
+		case <-ctx.Done():
+			close(jobs)
+			wg.Wait()
+			return
+		}
+	}
+	close(jobs)
+	wg.Wait()
+}
```


Tests for this would be nice too 😅

ArthurSens (commented):
I completely understand that `cfg.DataSourceNames[0]` is something that already exists in the current master branch. However, I think this is very related to the work here; you even mention in your PR description that you're trying to solve this problem, unless I'm misreading it 🤔 

If my understanding of the codebase is correct, years ago we tried to make database autodiscovery automatic but always failed because of that hardcoded `[0]`. What I'm saying is that it is finally time to change this field from an array of strings to a single string, keep the current behavior, and stop pretending we try to discover databases from all connections in that array.

`/metrics` (single target) and `/probe` (multi-target) endpoints both use the `Runtime` abstraction, and you're applying this fix on top of Runtime. This means both endpoints benefit from your fix! I believe this is a good move that fixes a very old problem we had :)

collector/pg_statio_user_indexes.go
sysadmind: Hoping to get back to a fresh review of this PR later in the week, but wanted to comment on this. Why does adding a label break users? Promql uses labels as filters, so a query for `my_metric{foo="bar"}` would match both `my_metric{foo="bar"}` and `my_metric{foo="bar",datname="baz"}` right? I think I may just be missing the use case that adding a label breaks. Do you have an example of what is breaking @ArthurSens? Whether or not it's breaking would change my opinion somewhat here, so probably good to understand that before making any decisions on direction.

collector/runtime.go
sysadmind: I think any decisions on `DataSourceNames` could be follow ups issues or PRs. We only use the first one today, so that is acceptable to me for this PR. I do think that having the ability to have server and database scoped metrics, like this PR adds, does mean that we could consider accepting multiple DSNs again, but we should decide that intentionally and in a separate change.

collector/pg_statio_user_indexes.go
ArthurSens: There are a couple of side effects:
* The same query will return more time series (depending on the cardinality of the new label). Dashboards will need to be reworked, and alerts might fire multiple times instead of just once
* Breaks vector matching of joins using `on(...)` or `ignoring(...)`
* Breaks aggregators using `without(...)`
* Prometheus has scrape label limits, which might trigger as new labels are added

ArthurSens (commented):
I'm taking another look at this while using some LLM tools to help me find potential bugs, and it found one that I'm struggling to propose a solution to.

When discovering databases, with a DSN like the one below:
```
host=db.example user=exporter password='pa\'ss' dbname=postgres
```

lib/pq will parse the password as `pa'ss`, which is the intended behavior. We do use lib/pq for the first connection, but for some reason we have a custom parser for all other databases that are not the primary. This custom implementation aparently does not use lib/pq for parsing and our implementation has side effects like not parsing escaped strings correctly or skipping emty strings (which is also valid for lib/pq). 

I spent some time trying to find solutions here but LLMs always came up with very complex solutions, which suggests it is not as easy to fix as I initially thought. Also, looking at the PR diff, the bug was not introduced now, it is something that existed even before.

ArthurSens:
And I'm so sorry for the previous guidance I gave you @mwimpelberg28, I've spent some extra time trying to understand if changing `DATA_SOURCE_NAME` from an array of strings to a single string is safe, and I believe it's not as safe as I initially thought.

Our codebase is currently split into two parts: a newish package that we call `collectors`, which has a better architecture for collecting metrics from Postgres and is the one we want to drive forward, and a legacy package called `exporter`, which we've been trying to deprecate and delete going forward. In my previous review, I completely forgot that `exporter` also depends on this variable, and the array is actually used properly there.

With DATA_SOURCE_NAME accepting several connections split by commas, we have this behavior today
```
DATA_SOURCE_NAME=[server-a/db1, server-b/db2]
                         │
            ┌────────────┴────────────┐
            │                         │
      pkg exporter        pkg collectors
            │                         │
       both DSNs                 first DSN only
            │                         │
 server-a + server-b              server-a
```

If we stop accepting multiple connections without notice, we'll break users who still depend on the legacy code path. With all that said, I think we need to go back to accepting the arrays and we work towards the deprecation in a separate PR. Sorry for the trouble 😬

README.md
```diff
@@ -297,6 +301,11 @@ The following environment variables configure the exporter:
 * `PG_EXPORTER_AUTO_DISCOVER_DATABASES` (DEPRECATED)
   Whether to discover the databases on a server dynamically. Value can be `true` or `false`. Default is `false`.
 
+* `PG_EXPORTER_AUTO_DISCOVER_DATABASES_MAX_CONCURRENCY`
+  Maximum number of discovered databases to scrape concurrently in a single scrape, when
+  `auto-discover-databases` is enabled. Bounds how many extra connections a scrape can open on a
+  server with many databases. Default is `10`.
```
ArthurSens: We're trying to deprecate all env vars that aren't required for security reasons. Could we avoid this one? :)

ArthurSens (commented):
Another test that I did with this PR was updating the dependency at https://github.com/prometheus/prometheus-opentelemetry-collector/tree/main/receivers/postgres and seeing what changes for downstream projects that use the `collectors` package as a library. I've noticed that now we have metrics from discovered databases, while we would only have for primary databases on current master :)

Some would call this a breaking change and others a bug fix. On my point of view it is a bug fix, and the project I linked above is considered experimental so it's fine to have behavior changes :)

ArthurSens (approved):
Ok, things are looking good from my side. It should be good to go after all commits are properly signed.

I would still like to see a final review from @sysadmind before merging though, it's a big change so maybe he will find some problems that I couldn't find :)

mwimpelberg28:
> Ok, things are looking good from my side. It should be good to go after all commits are properly signed.
> 
> I would still like to see a final review from @sysadmind before merging though, it's a big change so maybe he will find some problems that I couldn't find :)

Thanks Arthur!  I fixed the signing.

Issue #1386, opened 2026-09-06
IAM Auth Module


matt-gp:
## Proposal
Hi all. I'd like to look at adding an IAM Auth module. Given that multi-target is meant for one exporter to support multiple db's, I feel this would be a good edition. Currently I have to have a basic auth user on each database, which is fine for the minute, but we have issues with credential rotation etc. Having IAM would mean we would be able to have an IAM role which can talk to multiple the db's and the credentials would not need rotating.

**Note** this would only be for multi-target, previous attempts at this weren't specifically for multi-target, and hence had problems when the token expired after 15mins.

Thoughts?

sysadmind:
I think it would be great to add IAM auth in general. I would love to hear more about the credential rotation too - I think others could have similar struggles so it would be good to think about how we solve those too. From seeing your work on the AWS SD for Prometheus, I think you're well positioned to make the change. I think it will be a relatively small change, but if it's not, I would love to hear about the design before you get too far.

matt-gp:
Cool, I'll prepare a fork.

matt-gp:
@sysadmind when you say "add IAM auth in general" do you mean a generic IAM module that could be used with AWS/GCP/Azure. Or do you mean IAM which works for non multi-target as well?

sysadmind:
I mean IAM which works for multi-target and single target. I think based on the way we do auth, it would work for both depending on how you decide to do the implementation.

PR #1392, opened 2026-09-12
IAM Auth

matt-gp:
On the back of https://github.com/prometheus-community/postgres_exporter/issues/1386 this change adds IAM Auth.

IAM Auth allows RDS users to connect using dynamically generated tokens. The advantage of using this over username/password means that you don't need to rotate any credentials, so it is the preferred connection method when using RDS.

This change adds IAM Auth support to the multi-target (`/probe`) flow by creating an IAM Auth module. The change also adds IAM Auth support to the single target (`/metrics`) flow by allowing the configuration using environment variables.


matt-gp:
@sysadmind

internal/connector/awsiam.go
ArthurSens, line 1: I am very, very ignorant about IAM and AWS stuff, but I'm assuming that Prometheus itself supports IAM for SD authentication, and I'm also hoping that other exporters, such as Elasticsearch, Postgres, and others, will also need similar features.

Does it make sense to make this connector a reusable library?

If I'm talking bullshit, please don't hesitate to say so 😆 !

README.md
```diff
@@ -91,6 +91,41 @@ auth_modules:
       sslmode: disable
 ```
 
+#### IAM auth modules
```
ArthurSens, line 94: Joe and I are working to make exporter documentation available on the Prometheus website via https://github.com/prometheus-community/postgres_exporter/pull/1361.

Could we wait until that PR is merged and then update the docs accordingly?

README.md
```diff
@@ -314,6 +349,15 @@ The following environment variables configure the exporter:
...
 
+* `DATA_SOURCE_AUTH`
+  Set to `iam` to authenticate the primary datasource (`/metrics`) with AWS RDS/Aurora IAM database authentication instead of a static password, minting a fresh token for every connection. Requires the DB user (from `DATA_SOURCE_USER`) to have `rds_iam` granted and no password; `DATA_SOURCE_PASS`/`DATA_SOURCE_PASS_FILE` are ignored. The database name comes from `DATA_SOURCE_URI`/`DATA_SOURCE_NAME` as usual. Unset (the default) uses a static password as before.
+
+* `DATA_SOURCE_REGION`
+  Only used when `DATA_SOURCE_AUTH=iam`. The AWS region of the target cluster. Optional; if unset, resolved via the AWS SDK's usual chain (env, shared config, instance metadata).
+
+* `DATA_SOURCE_ROLE`
+  Only used when `DATA_SOURCE_AUTH=iam`. An IAM role ARN to assume via STS before minting a token. Optional; if unset, uses whatever AWS credentials are already available (e.g. an instance/task role).
+
```
ArthurSens, line 360: We're currently trying to deprecate and remove environment variables. We probably want to find a different way to configure IAM

ArthurSens (commented):
Didn't have time to do a proper review, I'll spend more time on this later :)

internal/connector/awsiam.go
matt-gp: Prometheus calls the AWS API for SD and the IAM role that is attached to Prometheus needs the correct perms. This effectively is slightly different though as this calls STS behind the scenes and gets it's own credentials.

RDS IAM Auth is a bit different as you generate a short lived token using an RDS API, and then use this token as the password. The db user that you are connecting as has to be part of a special group `rds_iam` and this means that it can auth using a short lived token.

Making it into a reusable library would be a good idea but I'm not sure how feasible it actually is. To generate the IAM token for RDS you need to call a specific RDS API, for elasticache this doesn't actually exist (https://github.com/aws/aws-cli/issues/9044).

README.md
matt-gp: Cool, would you like me to remove the docs from my change in that case?

README.md
matt-gp: Cool although they do come in handy when running things in ECS/Fargate which don't support configmaps. 

The environment variables here were only used for backwards compatibility it supports the auth module approach also.

I'm quite a big fan of the auth modules, although I think maybe standardizing the format across exporters would be good, the mysql one uses an ini file for example. If we did that then we might be able to have a shared library that exporters can use for interacting with the auth modules.

sysadmind: I wouldn't block this PR on the environment variables. The AWS SDK will also automatically load the ENV vars it needs when set to do so which is exactly what a user deploying in AWS would expect. The tradeoff to consider is that it would get confusing for multi-target. Would that case use ENV vars or auth modules or would one be favored over the other?

internal/connector/awsiam.go
matt-gp: Just thinking out loud but maybe creating an exporter toolkit library that could be where a load of shared modules live could be an option.

The rds iam auth even is applicable to both MySQL and postgresql. Whatever the elasticache equivalent is probably could be used on Redis and Memcached (and valkey). So would be a good opportunity to standardise a lot of things.

We could always work on that and migrate things over as/when.

sysadmind: We talked about this in the exporter maintainers call on Friday 9/18 - in this case I think the AWS code itself is really small here. I'm not that sure what we would gain from moving that specific code to another repo in this case.

internal/connector/awsiam.go
```diff
@@ -0,0 +1,114 @@
...
+// Driver implements the Connector interface.
+// It returns the underlying pq.Driver.
+func (c *AWSIAMConnector) Driver() driver.Driver {
+	return pq.Driver{}
+}
+
+// loadCredentials loads AWS credentials, optionally assuming a role if RoleARN is set.
+// It caches the credentials for reuse across Connect calls.
+func (c *AWSIAMConnector) loadCredentials(ctx context.Context) (aws.CredentialsProvider, error) {
+	cfg, err := awsconfig.LoadDefaultConfig(ctx, awsconfig.WithRegion(c.Region))
```
sysadmind, line 90: This should resolve region (if it's possible to), but auth.BuildAuthToken won't get it because it only sees c.Region which could still be an empty string. I think we need something like: 

```golang
cfg, err := awsconfig.LoadDefaultConfig(ctx, awsconfig.WithRegion(c.Region))
if err != nil {
    return nil, err
}
if c.Region == "" {
    if cfg.Region == "" {
        return nil, errors.New("no AWS region configured; set iam.region (or AWS_REGION)")
    }
    c.Region = cfg.Region
}
```

internal/connector/awsiam.go
```diff
@@ -0,0 +1,114 @@
...
+		Region:     region,
+		RoleARN:    roleARN,
+		Options:    options,
+	}
+}
+
+// Connect implements the Connector interface.
+// It mints a fresh IAM auth token on every connection attempt, optionally assuming a role via STS if RoleARN is set.
+func (c *AWSIAMConnector) Connect(ctx context.Context) (driver.Conn, error) {
+	c.once.Do(func() { c.creds, c.err = c.loadCredentials(ctx) })
```
sysadmind, line 64: Because this is a sync.Once, that means a failure is permanent too. Could we handle this another way?

internal/connector/awsiam.go
```diff
@@ -0,0 +1,114 @@
...
+	if c.err != nil {
+		return nil, fmt.Errorf("loading AWS credentials: %w", c.err)
+	}
+
+	token, err := auth.BuildAuthToken(ctx, c.DBEndpoint, c.Region, c.DBUser, c.creds)
```
sysadmind, line 69: Can you try this with a DSN where the port is missing? The docs suggest that port is required for BuildAuthToken to work.  https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/rds/auth#BuildAuthToken

Port is typically optional for our server endpoints so I think we should have some way to handle that as the input and return a nice user-friendly error.

internal/connector/awsiam.go
```diff
@@ -0,0 +1,114 @@
...
+		return nil, err
+	}
+	if c.RoleARN == "" {
+		return cfg.Credentials, nil
+	}
+	return aws.NewCredentialsCache(stscreds.NewAssumeRoleProvider(sts.NewFromConfig(cfg), c.RoleARN)), nil
+}
+
+// dsn builds the connection string for a single connection attempt, with the freshly minted token as the password.
+func (c *AWSIAMConnector) dsn(token string) string {
```
sysadmind, line 101: I worry a bit that we're duplicating setting up the connection here and that could diverge over time unintentionally. Do you think that we could adapt collector.instance instead?

config/config.go
```diff
@@ -357,3 +369,53 @@ func (m AuthModule) ConfigureTarget(target string) (DSN, error) {
...
+		// options override target's own query parameters when set, same as for userpass.
+		options := make(map[string]string, len(dsn.query)+len(m.Options))
+		for k := range dsn.query {
+			options[k] = dsn.query.Get(k)
+		}
+		for k, v := range m.Options {
+			options[k] = v
+		}
+
+		return connector.NewAWSIAMConnector(
```
sysadmind, line 406: This means that we totally rebuild the AWS auth including IMDS and STS on every request to /probe right? That feels wasteful. Is there a way to adapt the auth module to hold onto that information?

collector/runtime.go
```diff
@@ -29,7 +30,7 @@ type Runtime struct {
 	postgresCollector *PostgresCollector
 }
 
-func NewRuntime(validatedConfig config.ValidatedConfig, logger *slog.Logger) (*Runtime, error) {
+func NewRuntime(validatedConfig config.ValidatedConfig, logger *slog.Logger, conn driver.Connector) (*Runtime, error) {
```
sysadmind, line 33: I think this is a breaking change for the OTEL implementation right @ArthurSens?

internal/connector/static_test.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2021 The Prometheus Authors
```
sysadmind, line 1:
```suggestion
// Copyright The Prometheus Authors
```

internal/connector/static.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2021 The Prometheus Authors
```
sysadmind, line 1:
```suggestion
// Copyright The Prometheus Authors
```

internal/connector/awsiam_test.go
```diff
@@ -0,0 +1,107 @@
+// Copyright 2021 The Prometheus Authors
```
sysadmind, line 1:
```suggestion
// Copyright The Prometheus Authors
```

internal/connector/awsiam.go
```diff
@@ -0,0 +1,114 @@
+// Copyright 2021 The Prometheus Authors
```
sysadmind, line 1:
```suggestion
// Copyright The Prometheus Authors
```

sysadmind (changes requested):
I think the connector is a reasonable direction - my research indicates that it's common. I want to make sure that we're being careful about diverging from our existing connection setup though. It buys us a lot like the fact that it will redact credentials in errors while still being able to log the DSN.

collector/runtime.go
ArthurSens: It is, but we don't promise API stability just yet. We can handle it downstream :) 

That said, if we can work around it without breaking the API, that would be awesome. I haven't tested yet how these changes affect us downstream; it's on my TODO list to test things out

README.md
ArthurSens: The docs PR got merged, I believe we can move things there now :)

collector/runtime.go
ArthurSens: Maybe we could do `NewRuntime(config, logger, ...RuntimeOptions)` here?

PR #1393, opened 2026-09-18
Bound legacy queries by collection timeout

maltere:
## Problem

The legacy exporter path used by `--extend.query-path` executes database operations without a context and is not covered by `--collection-timeout`.

If a custom query becomes slow or blocks, an HTTP scrape timeout does not stop the underlying database work. Repeated scrapes can therefore accumulate blocked collection goroutines and associated resources.

## Change

- Pass the existing collection timeout to the legacy exporter.
- Create a timeout context for each legacy collection.
- Use context-aware database operations for connection checks, version detection, database discovery, and metric queries.
- Add a test confirming that a slow custom query stops when its context expires.

## Testing

```text
go test ./...
```


maltere:
@sysadmind

exporter/namespace.go
ArthurSens, line 184: I think we need to check for rows.Err() here now. If the context ends before the rows finished being scanned, an error will be returned and we need to check it

ArthurSens: And thinking a bit more, maybe we want to start exposing metrics after we scan everything. We don't want to expose a partial scrape 😬

exporter/server.go
ArthurSens, line 200: now that we have context, maybe we could get rid of this sleep?

I'm just thinking out loud, to be honest; I'm not sure how this could be done... but it sounds weird to sleep when the timeout hits... this means our collection always takes longer than the actual timeout 🤔

ArthurSens (commented):
Thanks for the contribution! I've added a few comments below :)

exporter/namespace.go
maltere: Yes good catch. Will add this :)

exporter/server.go
maltere: Hm sounds right, I have replaced both unconditional `time.Sleep` calls with context-aware waits. The delay now exits immediately when `ctx.Done()` is triggered, so collection does not continue beyond its timeout.

exporter/queries.go
ArthurSens, line 138: same thing here, we also need to check rows.Err() now

ArthurSens (commented):
just noticed one thing we missed, otherwise LGTM :)

exporter/queries.go
maltere: Added!

ArthurSens (approved):
Thanks!

